Issue #39, opened 2017-01-19
Undelegated test (CLI) ignores name that does not resolve to address


matsduf:
As @aabdnn writes in issue https://github.com/dotse/zonemaster-engine/issues/278 there is a bug in Zonemaster-CLI when an undelegated test is fed with name that cannot be resolved to an address:

> I expect this test to fail (because name server abcdef does not exist):
> 
> ```
> zonemaster-cli 193.in-addr.arpa --ns pri.authdns.ripe.net --ns sns-pb.isc.org --ns tinnie.arin.net --ns sec3.apnic.net --ns abcdef
> ```
> 
> However, Zonemaster (CLI 1.0.5, engine 1.0.16) happily passes this, with no errors emitted.

As far as I can see, method `add_fake_delegation` in `Zonemaster/CLI.pm` just ignores any name that cannot be resolved to an address.

```
        if ($ip) {
            push @{ $data{ $self->to_idn( $name ) } }, $ip;
        }
        else {
            my $n = $self->to_idn( $name );
            my @ips = Net::LDNS->new->name2addr($n);
            push @{ $data{$n} }, $_ for @ips;
        }
```

If name is fed without address, the address is looked up with `my @ips = Net::LDNS->new->name2addr($n);` and if that returns nothing, the name is ignored.

vlevigneron:
Fixed in https://github.com/dotse/zonemaster-cli/pull/41

PR #41, opened 2017-02-02
- Fixes dotse/zonemaster-cli#40

vlevigneron:


lib/Zonemaster/CLI.pm
```diff
@@ -517,13 +518,24 @@ sub add_fake_delegation {
         if ($ip) {
             push @{ $data{ $self->to_idn( $name ) } }, $ip;
         }
+        elsif (Zonemaster::Zone->new( { name => $domain } )->is_in_zone( $name )) {
+            say STDERR "--ns in zone nameservers must be a name/ip pair.";
+            exit( 1 );
+        }
```
matsduf: I do not think this is the correct design. The error should be like any other error. It should be caught be a test case and reported. With the solution above,  Zonemaster will not check the other nameservers provided.

lib/Zonemaster/CLI.pm
```diff
@@ -517,13 +518,24 @@ sub add_fake_delegation {
...
             my $n = $self->to_idn( $name );
             my @ips = Net::LDNS->new->name2addr($n);
-            push @{ $data{$n} }, $_ for @ips;
+            if ( @ips ) {
+                push @{ $data{$n} }, $_ for @ips;
+            }
+            else {
+                say STDERR "--ns ${name} nameserver has no IP address.";
+                exit( 1 );
+            }
```
matsduf: I do not think this is the correct design. The error should be like any other error. It should be caught be a test case and reported. With the solution above, Zonemaster will not check the other nameservers provided.

lib/Zonemaster/CLI.pm
vlevigneron: In facts, this error is like others... CLI errors. It is consistent with the way the CLI was written. We can plan improvements. First we can check all args and just exit when all args has been verified, not only when it meets the first mistake. This work is only on the CLI part. Second, we can add this in the engine itself, it is more complicated because, as you noticed, the design must be modified.

If the user demand is to catch cases when a nameserver is not correct or need an IP because it is in zone, the first solution can be enough, at least at a first step.

What do you thing @aabdnn ?

lib/Zonemaster/CLI.pm
aabdnn: I agree with @matsduf. There are 2 types of checks: syntactic and semantic. The cli (and the backend) should ONLY do syntactic checks. They should ensure that the test being requested has the correct syntax. So with a name server, the cli should just check that a paramater has been supplied to the --ns option.

Then the cli should hand over the test to the engine, which is responsible for SEMANTIC checks, such as:
1. Is the given name server name a valid DNS label?
2. If it is in-zone, was an IP address also provided?

The point I am making is that the CLI should do as little as possible. It should not resolve names to addresses itself. The CLI is just a wrapper that interfaces to the engine. In all cases, the engine should do all the testing and report any errors.

Similarly, the backend's API is just an interface to the engine. It should only check the syntax of the JSON request, and then pass it to the engine. Any problems with the name server or resolving it to addresses, etc, should be checked for, and reported by, the engine.

The current model is in my opinion wrong. The cli (and the backend) are doing some of the work themselves, and hiding errors from the engine. But this is also inconsistent, and confusing. I should be able to submit a check via the CLI or the backend, and expect the exact same result from the engine. And the only way for this to happen is for both the CLI and backend to just be wrappers for the engine, and not do any of the DNS checks themselves.

aabdnn: Actually, this issue highlights a design problem, rather than a bug, in the cli (and the backend). Can you guys bring this to the attention of the entire dev team, so that everyone is on the same page? And then see how you can rework both the cli and backend to do the right thing and be consistent.

vlevigneron:
OK, my proposal was a response to the issue
https://github.com/dotse/zonemaster-cli/issues/39 that was supposidely
blocking you and needed a quick fix. I understand now you need a fix 
for both CLI and Backend and that you can wait for a more complete
solution.

If we go, as I proposed, to plan B (modify engine behaviour for "fake
delegations"), it will take more time because, as you noticed, it as not
been designed for that kind of case.

le 03 févr., Anand Buddhdev a ?crit :
> aabdnn commented on this pull request.
> 
> 
> 
> >          else {
>              my $n = $self->to_idn( $name );
>              my @ips = Net::LDNS->new->name2addr($n);
> -            push @{ $data{$n} }, $_ for @ips;
> +            if ( @ips ) {
> +                push @{ $data{$n} }, $_ for @ips;
> +            }
> +            else {
> +                say STDERR "--ns ${name} nameserver has no IP address.";
> +                exit( 1 );
> +            }
> 
> Actually, this issue highlights a design problem, rather than a bug, in the cli (and the backend). Can you guys bring this to the attention of the entire dev team, so that everyone is on the same page? And then see how you can rework both the cli and backend to do the right thing and be consistent.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/dotse/zonemaster-cli/pull/41

-- 
	Vincent Levigneron  A.F.N.I.C.  Vincent.Levigneron@nic.fr

aabdnn:
@vlevigneron issue #39 highlights the problem, but we first discovered it in the backend. That's because the backend is used for all checks of reverse DNS delegations for the RIPE Database, and we don't want users getting away with non-existent name servers. When my colleague showed me the issue with the backend, I use the cli to see the problem. I think the cli and backend both need fixes, but the engine also needs a fix.

vlevigneron:
With  https://github.com/dotse/zonemaster-engine/pull/285, this patch is no longer relevant. A new proposal will be done according Engine changes.

PR #44, opened 2017-06-03
- Fixes dotse/zonemaster-cli#43

vlevigneron:
@matsduf @mattias-p Need a review for name space changes

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -24,16 +24,8 @@ configure_requires( 'Module::Install' => 0, );
 
 build_requires( 'Test::More' => 0, );
 
-# Dirty hack to avoid complaints about unparsable version number from
-# old versions of ExtUtils::MakeMaker.
-if ( eval { ExtUtils::MakeMaker->VERSION(6.68) } ) {
-    requires 'Zonemaster::Engine' => 'v1.0.4';
-}
-else {
-    requires 'Zonemaster::Engine' => 0;
-}
-
 requires(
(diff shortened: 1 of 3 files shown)


Makefile.PL
```diff
@@ -27,10 +27,10 @@ build_requires( 'Test::More' => 0, );
 # Dirty hack to avoid complaints about unparsable version number from
 # old versions of ExtUtils::MakeMaker.
 if ( eval { ExtUtils::MakeMaker->VERSION(6.68) } ) {
-    requires 'Zonemaster' => 'v1.0.4';
+    requires 'Zonemaster::Engine' => 'v1.0.4';
 }
 else {
-    requires 'Zonemaster' => 0;
+    requires 'Zonemaster::Engine' => 0;
```
matsduf: 1. Is the "dirty hack" really needed? Isn't 1.000004 = v1.0.4 in the perl world?
2. I think we should use the real version of Zonemaster Engine, at least at the time of release. I guess it will be v2.0.0.

matsduf:
Else I see no problem with the PR. I can install and run it together with the update Zonemaster Engine. 

I notice that there is only one unit test for Zonemaste CLI. I think there are more things worth testing, e.g. locale switching through the command line option.

vlevigneron:
@matsduf  You are right, we should add more unit tests. I'll add some in the future.

matsduf:
@vlevigneron, I approve the PR. You can merge.

PR #51, opened 2017-10-19
Net::LDNS has been renamed to Zonemaster::LDNS

mattias-p:


mattias-p:
@matsduf, I agree that the .mo files don't belong in the repo but I think that's a separate concern that belongs in a separate PR.

PR #52, opened 2017-10-19
Various updates to packaging

mattias-p:


share/Makefile
```diff
@@ -4,12 +4,8 @@ MOFILES := $(POFILES:.po=.mo)
 MSGFMT  = msgfmt --check
```
matsduf: I think we should switch to Perl msgfmt instead of binary.

mattias-p: Done

PR #57, opened 2017-10-27
Update Makefile.PL

matsduf:
Specify smallest version of Locale::TextDomain.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -23,7 +23,7 @@ build_requires( 'Test::More' => 0, );
 
 requires(
     'JSON::XS'           => 0,
-    'Locale::TextDomain' => 0,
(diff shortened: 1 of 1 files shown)


Makefile.PL
```diff
@@ -23,7 +23,7 @@ build_requires( 'Test::More' => 0, );
 
 requires(
     'JSON::XS'           => 0,
-    'Locale::TextDomain' => 0,
+    'Locale::TextDomain' => 1.23,
```
mattias-p: Shouldn't it be 1.21?

matsduf: 1.23 is the lowest version that we have verified. As our instructions are written, everybody will get 1.28 (or higher).

PR #58, opened 2017-10-31
Update licensing

mattias-p:
* Update full name of IIS
* Specify licensing for documentation
* Remove conflicting license reference
* Update years of copyright

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -12,7 +12,6 @@ all_from 'lib/Zonemaster/CLI.pm';
 tests_recursive( 't' );
 
 resources(
-    license    => 'http://www.opensource.org/licenses/BSD-3-Clause',
     repository => 'https://github.com/dotse/zonemaster-cli',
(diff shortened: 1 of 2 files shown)


Makefile.PL
```diff
@@ -12,7 +12,6 @@ all_from 'lib/Zonemaster/CLI.pm';
 tests_recursive( 't' );
 
 resources(
-    license    => 'http://www.opensource.org/licenses/BSD-3-Clause',
```
matsduf: Shouldn't it rather be

license => 'https://opensource.org/licenses/BSD-2-Clause'

mattias-p: No, the value of the resources/license property is overridden by a value derived from the license property when META.yml is generated. When license is set to "bsd" the URL that ends up in META.yml becomes "http://opensource.org/licenses/bsd-license.php". That URL points to a BSD-2 template.

PR #59, opened 2017-11-01
Preparation for release of v1.1.1.

matsduf:


Changes
```diff
@@ -1,4 +1,17 @@
-Release history for Perl module Zonemaster::CLI
+Release history for Zonemaster Component Zonemaster-CLI 
+(Perl module Zonemaster::CLI).
```
mattias-p: This changelog is not limited to just that module. The Zonemaster-CLI distribution would be more accurate. Alternatively, I believe this line could be removed without loss of information.

Changes
```diff
@@ -1,4 +1,17 @@
-Release history for Perl module Zonemaster::CLI
+Release history for Zonemaster Component Zonemaster-CLI
```
mattias-p: Should component really be spelled with a capital c?

Changes
```diff
@@ -1,4 +1,17 @@
...
+- Changed dependency from Zonemaster to Zonemaster::Engine due
+  to name space change (#43)
+- Fix Commonmark rendering on Github. Replace NBSP with SPACE. (#49)
+- Various updates to packaging (#52) 
+- Changed dependency from Net::LDNS to Zonemaster::LDNS due to
+  name space change (#51) 
+- Updated installation instruction (#53) 
+- Specify smallest version of Locale::TextDomain i Makefile.PL. (#57)
+- Update licensing (#58) 
+
```
mattias-p: Maybe add a heading indicating that all these things have been "Fixed" to associate them with the patch level?

I was thinking about suggesting headings for Changed, Removed (major), Added (minor), Fixed, Deprecated (patch), but I never got to it until now.

PR #75, opened 2018-05-11
Updated the install instructions for debian and centos

sandoche2k:


mattias-p:
This was supposed to go against the develop branch, right?

matsduf:
Yes. I did not see that when I approved. Is reverting and adding it to the develop branch the right thing? Or just adding it to the develop branch?

mattias-p:
The important thing is that the change gets added to the develop branch.
Reverting it on master is the proper thing, but looking at the change it shouldn't make much difference in practice.

Issue #77, opened 2018-05-28
Print diagnostics to STDERR


mattias-p:
The `--config` and `--policy` parameters print diagnostic information about what file was loaded (e.g. "Loading policy from myfile.json"). Today these diagnostics are printed to STDOUT. This behavior is problematic when the output is fed to `jq` for pretty printing. The diagnostics should be printed to STDERR instead in order to keep them separated from the normal output.

matsduf:
PR #84 has been created.

mattias-p:
PR #84 solved some aspects of this issue, but it did not resolve the problem with regard to the `--json_stream` option.

mattias-p:
Fixed in #85.

PR #85, opened 2018-08-20
Diagnostics

mattias-p:
Solves #77. It also includes collateral improvements regarding L10N.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -291,15 +291,18 @@ sub run {
         Zonemaster::Engine->config->resolver_source($self->sourceaddr);
     }
 
+    # Filehandle for diagnostics output
+    my $fh_diag = ( $self->json or $self->json_stream or $self->raw or $self->dump_config or $self->dump_policy )
+      ? *STDERR     # Structured output mode (e.g. JSON)
+      : *STDOUT;    # Human readable output mode
+
     if ( $self->policy ) {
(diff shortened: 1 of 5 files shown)


lib/Zonemaster/CLI.pm
```diff
@@ -291,15 +291,18 @@ sub run {
         Zonemaster::Engine->config->resolver_source($self->sourceaddr);
     }
 
+    # Filehandle for diagnostics output
+    my $fh_diag = ( $self->json or $self->json_stream or $self->raw or $self->dump_config or $self->dump_policy )
+      ? *STDERR     # Structured output mode (e.g. JSON)
+      : *STDOUT;    # Human readable output mode
+
```
matsduf: I see "raw" as still human readable mode. It is not a defined format. It is the same format as the "standard format". It is just a way to avoid the translation and to get the messages directly. Move "raw" to the STDOUT. 

Otherwise I like your structuring of the code. It makes it more consistent and easier to maintain.

mattias-p: I'm glad you like the structuring. There's still room for improvement, even in this area of the code, but I think these changes are enough for now.

Raw mode is structured enough for processing with `grep -q` or `sed`. Doing so is simpler than using `jq` and that is useful in many situations. Moreover combining STDERR and STDOUT with the good old "2>&1" is simpler than separating diagnostics from normal output with standard tools. I think these arguments are enough to categorize raw mode as a structured output mode.

PR #92, opened 2018-11-07
Test against current Engine develop branch 

mattias-p:
Test against current Zonemaster::Engine develop branch instead of the latest version of Zonemaster::Engine according to CPAN.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,8 @@ perl:
     - "5.18"
     - "5.16"
     - "5.14"
+before_install:
+    - eval $(curl https://travis-perl.github.io/init) --auto
+    - local-lib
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,3 +6,8 @@ perl:
     - "5.18"
     - "5.16"
     - "5.14"
+before_install:
+    - eval $(curl https://travis-perl.github.io/init) --auto
+    - local-lib
+    - git clone --depth=1 --branch=develop https://github.com/zonemaster/zonemaster-engine.git
+    - cpan-install --deps ./zonemaster-engine
```
matsduf: I have no objections but one question. Wouldn't downloading the zip file with the checkout files in the develop branch giving less overhead to Travis and possibly being faster? The Git structure is of no importance here.

https://github.com/zonemaster/zonemaster-engine/archive/develop.zip

PR #93, opened 2018-11-09
Profiles

vlevigneron:
Works with corresponding zonemaster::Engine profiles branch.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -6,7 +6,7 @@ extends 'Zonemaster::Engine::Exception';
 # The actual interesting module.
 package Zonemaster::CLI;
 
-use version; our $VERSION = version->declare("v1.1.3");
+use version; our $VERSION = version->declare("v1.2.0");
 
 use 5.014002;
 use warnings;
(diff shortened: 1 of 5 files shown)


matsduf:
@vlevigneron, I would have expected updated CLI to fail against current Engine. Do we lack some unit tests?

vlevigneron:
le 09 nov., Mats Dufberg a ?crit :
> @vlevigneron, I would have expected updated CLI to fail against current Engine. Do we lack some unit tests?
> 

Perhaps, I have not tested it against current version of Engine, only
updated one.

> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/zonemaster/zonemaster-cli/pull/93#issuecomment-437299725

-- 
	Vincent Levigneron  A.F.N.I.C.  Vincent.Levigneron@afnic.fr

lib/Zonemaster/CLI.pm
```diff
@@ -292,37 +307,47 @@ sub run {
...
+        Zonemaster::Engine->profile->load( $self->profile );
     }
+    else {
 
-    if ( $self->config ) {
-        say $fh_diag __x( "Loading configuaration from {path}.", path => $self->config );
-        Zonemaster::Engine->config->load_config_file( $self->config );
+        if ( $self->policy ) {
+            say $fh_diag __x( "Loading policy from {path}.", path => $self->policy );
+            say $fh_diag __x( "DEPRECATED, use profile instead." );
```
mattias-p: We should use a stronger word than deprecated if we make it an error to use it. Deprecation implies that the old way still works for the time being but will be removed in a future version.
https://en.wikipedia.org/wiki/Deprecation

matsduf: I suggest that we write "Use of config and policy have been TERMINATED. Use profile instead."

matsduf:
@vlevigneron, Travis has done the tests against the current version. I think it means that it lacks some unit tests when it passes.

lib/Zonemaster/CLI.pm
vlevigneron: Done this way.

vlevigneron:
@matsduf Theer are very few UT in the CLI, compared to the Engine. I agree we should improve this in the future.

sandoche2k:
@matsduf and @mattias-p After the discussion last week.. do you agree to accept this PR waiting for updating the travis test requirements in the future?

sandoche2k:
@mattias-p thanks for reviewing, so that @vlevigneron can merge

PR #95, opened 2019-01-08
Update .travis.yml

matsduf:
Ubuntu 18.04 uses Perl 5.26. Default perl in FreeBSD is 5.28. No support OS uses perl 5.14.


matsduf:
Why does not Travis start here?

PR #105, opened 2019-05-28
Update Makefile.PL

matsduf:
Updated with correct version on dependencies.


matsduf:
Solves the same problem as in Zonemaster-Backend. See zonemaster/zonemaster-backend#518

Issue #38, opened 2016-12-28
Less helpful error message when source address of --sourceaddr cannot be used


matsduf:
Zonemaster CLI has the option --sourceaddr. If that is set to localhost or to an adress not bound to any interface, the error message is less helpful.

```
$ ifconfig eth0 | grep "inet "
          inet addr:109.74.12.66  Bcast:109.74.13.255  Mask:255.255.254.0

$ zonemaster-cli --sourceaddr 109.74.12.67 google.de
Seconds Level     Message
======= ========= =======
   0.15 ERROR     No response from nameserver for zone . when trying to fetch glue.
   0.17 CRITICAL  Not enough data about google.de was found to be able to run tests.

$ zonemaster-cli --sourceaddr 127.0.0.1 google.de
Seconds Level     Message
======= ========= =======
   0.14 ERROR     No response from nameserver for zone . when trying to fetch glue.
   0.15 CRITICAL  Not enough data about google.de was found to be able to run tests.

$ zonemaster-cli --sourceaddr ::1 google.de
Seconds Level     Message
======= ========= =======
  90.32 ERROR     No response from nameserver for zone . when trying to fetch glue.
  90.34 CRITICAL  Not enough data about google.de was found to be able to run tests.
```

**NOTE** (2021-01-19): #110 fixes the first case (using a address not attached to any interface) but not the second and third (using a localhost address). The first case was the most important, so this issue is considered to be resolved.

sandoche2k:
@matsduf can you propose a text how the error message should be ?

matsduf:
"Address ADDR cannot be used as source address for DNS queries."

`zonemaster-cli` should stop at such error.

mattias-p:
This is an input validation issue for the command line interface.

matsduf:
PR created in #110.

matsduf:
Resolved by #110.

PR #110, opened 2019-07-20
Fix for --sourceaddr bad behaviour

vlevigneron:
fixes zonemaster/zonemaster-cli#38
danish and swedish translation missing


lib/Zonemaster/CLI.pm
```diff
@@ -312,7 +315,12 @@ sub run {
     Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
 
     if ($self->sourceaddr) {
-        Zonemaster::Engine::Profile->effective->set( q{resolver.source}, $self->sourceaddr );
+        if ($self->check_sourceaddress_exists ) {
+            Zonemaster::Engine::Profile->effective->set( q{resolver.source}, $self->sourceaddr );
+        }
+	else {
+            die __( "Address '" ) . $self->sourceaddr . __("' cannot be used as source address for DNS queries.") . "\n";
```
mattias-p: The gettext documentation on [preparing translatable strings](https://www.gnu.org/software/gettext/manual/html_node/Preparing-Strings.html#Preparing-Strings) states that "[T]ranslatable strings should be entire sentences", and prescribes format strings as a solution. For consistency with existing translation strings, I suggest using [perl-brace-format](https://www.gnu.org/software/gettext/manual/html_node/perl_002dformat.html#perl_002dformat) (via `__x`). 

I recommend reading the entire page on preparing translatable strings. It's not too long and it provides good understanding of the interface between programmers and translators.

Makefile.PL
```diff
@@ -21,6 +21,7 @@ build_requires( 'Test::More' => 0, );
 requires(
     'JSON::XS'           => 0,
     'Locale::TextDomain' => 1.23,
+    'Net::Interface'     => 0,
```
mattias-p: I can't find a binary package for this on CentOS. Have you checked for alternatives?

lib/Zonemaster/CLI.pm
```diff
@@ -554,6 +562,32 @@ sub run {
...
+    my ( $self ) = @_;
+    my $address = Zonemaster::Engine::Net::IP->new($self->sourceaddr);
+    my $exists = 0;
+    foreach my $if ( Net::Interface->interfaces() ) {
+        foreach my $family ( AF_INET, AF_INET6 ) {
+            foreach my $ifaddr ( $if->address($family) ) {
+                my $zm_ifaddr;
+                if ( $family == AF_INET ) {
+                    $zm_ifaddr = Zonemaster::Engine::Net::IP->new(Net::Interface::inet_ntoa($ifaddr));
+	        }
```
mattias-p: Could you replace the tab with spaces?

lib/Zonemaster/CLI.pm
matsduf: I find [preparing translatable strings] worth reading and I agree that we should follow that advice.

[preparing translatable strings]: https://www.gnu.org/software/gettext/manual/html_node/Preparing-Strings.html#Preparing-Strings

matsduf:
@vlevigneron, what is the milestone of this PR? v2019.2 or v2019.3?

lib/Zonemaster/CLI.pm
vlevigneron: Fixed

lib/Zonemaster/CLI.pm
vlevigneron: Fixed

Makefile.PL
vlevigneron: @mattias-p  I only found Net::Interface that can handle with IPv4 and IPv6. IO::Interface for example does not work with IPv6. Any advices for an alternate library ?

matsduf: At least for CentOS 7 I find a package: https://centos.pkgs.org/7/openfusion-x86_64/perl-Net-Interface-1.012-2.of.el7.x86_64.rpm.html

vlevigneron: According to https://rt.cpan.org/Public/Bug/Display.html?id=124582, me should add this somewhere in travis.yml
`mkdir /usr/include/sys/; ln -s /usr/include/bits/socket.h /usr/include/sys/socket.h` to prevent the fail on Travis.

matsduf (dismissed):

Makefile.PL
mattias-p: @vlevigneron Thanks for the clarification regarding the choice of library. And no, I don't have any alternatives up my sleeve.

mattias-p:
Should the installation instruction be updated with the new dependency in this PR as well, or do we do that as part of installation testing?

vlevigneron:
@mattias-p Good point. I guess that it is usually done during packaging process, can anyone confirm ? If this is not the case, it can be modified in this PR.
@matts I finally fixed the issue I had with Travis, it has passed all the checks. Please approve again.

matsduf:
> @mattias-p Good point. I guess that it is usually done during packaging process, can anyone confirm ? If this is not the case, it can be modified in this PR.

If possible it should be done now. For FreeBSD it is p5-Net-Interface to be installed. CentOS has already CPAN dependencies so I guess it is OK to add it there. Fore Debian/Ubuntu we should have binary package.

> @Matts I finally fixed the issue I had with Travis, it has passed all the checks. Please approve again.

Wrong "(a)Matts". I will review again.

Matts:
You are not the first, and most certainly will not be the last that accidentally mentions me :P

Own fault for having a common name lmao

matsduf:
> You are not the first, and most certainly will not be the last that accidentally mentions me :P
> 
> My own fault for having a common name lmao. Besides, it is interesting to have a look at what other people are up to

@Matts, you were early to get a short user name. :-) You are welcome to Zonemaster!

share/da.po
```diff
@@ -131,6 +131,10 @@ msgid ""
 "Local IP address that the test engine should try to send its requests from."
 msgstr "Lokal IP-adresse som test-motoren skal sende sin foresgpørgsler fra."
 
+#, perl-brace-format
+msgid "Address {address} cannot be used as source address for DNS queries.\n"
+msgstr "Address {address} cannot be used as source address for DNS queries.\n"
+
```
mattias-p, line 137: I'm sorry I didn't spot this earlier, but I don't think this should be included. It only makes it difficult for the Danish translators to notice that there is something new for them to translate.

share/nb.po
```diff
@@ -167,6 +167,10 @@ msgstr ""
 "installert på dette system?\n"
 "\n"
 
+#, perl-brace-format
+msgid "Address {address} cannot be used as source address for DNS queries.\n"
+msgstr "Address {address} cannot be used as source address for DNS queries.\n"
+
```
mattias-p, line 173: This should also be excluded for the same reason.

matsduf (dismissed):

mattias-p (commented):
Everything looks good to me, except two problems I should have spotted earlier. Sorry about that.

share/nb.po
vlevigneron: @matsduf As you asked for something different (but it was a long time ago), do you think that @mattias-p proposal is reasonable ?

matsduf: That is fine. I am not sure if it should be empty or non-existent, but at least the latter will mean that msgid will shown, and then when you translate you will be prompted.

matsduf (approved):

mattias-p (approved):

Issue #111, opened 2019-07-29
DEBUG3 gives cluttered output


matsduf:
Up to and including DEBUG2 messages are single line. DEBUG3 includes something that looks like dig output, which is multiline. That make reading the output hard. Next is an excerpt from the following command, `zonemaster-cli --level=DEBUG3 --test=CONSISTENCY/consistency05 --show_module iis.se`:

```
   0.09 DEBUG2    SYSTEM       SYSTEM:NS_CREATED ip=202.12.27.33; name=m.root-servers.net
   0.10 DEBUG2    SYSTEM       SYSTEM:NS_CREATED ip=2001:0dc3:0000:0000:0000:0000:0000:0035; name=m.root-servers.net
   0.10 DEBUG2    SYSTEM       SYSTEM:RECURSE_QUERY address=2001:0dc3:0000:0000:0000:0000:0000:0035; class=IN; name=iis.se; ns=m.root-servers.net; source=m.root-servers.net/2001:dc3::35; type=SOA
   0.10 DEBUG2    SYSTEM       SYSTEM:QUERY flags={"class":"IN"}; ip=2001:dc3::35; name=iis.se; type=SOA
   0.10 DEBUG2    SYSTEM       SYSTEM:CACHE_CREATED ip=2001:0dc3:0000:0000:0000:0000:0000:0035
   0.11 DEBUG2    SYSTEM       SYSTEM:CACHE_FETCHED ip=2001:0dc3:0000:0000:0000:0000:0000:0035
   0.11 DEBUG     SYSTEM       SYSTEM:EXTERNAL_QUERY flags={"class":"IN"}; ip=2001:dc3::35; name=iis.se; type=SOA
   0.15 DEBUG3    SYSTEM       SYSTEM:EXTERNAL_RESPONSE packet=;; ->>HEADER<<- opcode: QUERY, rcode: NOERROR, id: 55847
;; flags: qr ; QUERY: 1, ANSWER: 0, AUTHORITY: 10, ADDITIONAL: 15 
;; QUESTION SECTION:
;; iis.se.      IN      SOA

;; ANSWER SECTION:

;; AUTHORITY SECTION:
se.     172800  IN      NS      z.ns.se.
se.     172800  IN      NS      x.ns.se.
se.     172800  IN      NS      y.ns.se.
se.     172800  IN      NS      j.ns.se.
se.     172800  IN      NS      a.ns.se.
se.     172800  IN      NS      c.ns.se.
[...]

PR #113, opened 2019-08-08
Add headers to every line of multiline log entries

mattias-p:
Add headers to all of the lines in `--raw` mode. To tell where new entries begin I added a marker character right after MODULE:TAG. The first line of an entry is marked with an extra SPACE character. The following lines are marked with a '>' character.

When you have headers on all lines you can filter the logs with better precision using grep.

**Edit:** I removed the reference to #111 as this PR solves a different problem.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -444,10 +444,16 @@ sub run {
                     if ( $self->show_testcase ) {
                         $prefix .= sprintf "%-14s ", $entry->testcase;
                     }
+                    $prefix .= $entry->tag;
 
                     my $message = $entry->string;
-                    $message =~ s/^([A-Z0-9]+:)*//;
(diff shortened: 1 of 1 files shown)


mattias-p:
@matsduf `--raw` is already prefixing some of the lines with headings. This change makes it so all lines are uniformly prefixed. I don't agree that it makes the output more difficult on the eye. In fact I argue the opposite. It also make the output more predictable to parse for machines. Please reconsider.

matsduf:
> Please reconsider.

I will check your branch again to see how it looks like in different cases.

mattias-p:
I had to rework this PR because the output format of the --raw mode had change, but the effect of the PR is essentially the same.

Please re-review.

matsduf (approved):
I have no specific use case to test against, but it looks OK. The output format of "raw" should be documented in a different PR.

PR #116, opened 2019-09-24
Updated Travis config after zonemaster-ldns#51

mattias-p:
zonemaster/zonemaster-ldns#51 adds feature checks for Ed25519, but Travis provides an older version of OpenSSL.


mattias-p:
I believe I've fully pinned this one down now. Please review.

PR #119, opened 2019-10-18
Centos installation

### LINKED ISSUE #117: Update installation instructions for CLI for CentOS 8
CentOS 8 is a new OS planned to be supported by release v2019.2. The installation instructions probably need to be adjusted for CentOS 8. There could be more binary packages available, but there could also be other changes.

mattias-p:
Fixes #117.

Also cleans up a useless explicit dependency on a core module.


matsduf:
This looks fine, but travis is unhappy.

PR #129, opened 2020-01-07
Removed all dependencies installation as they are not needed

sandoche2k:


sandoche2k:
@matsduf @mattias-p we had lot of discussion on this topic. I would like to get the confoirmation

docs/Installation.md
```diff
@@ -38,18 +38,6 @@ This instruction covers the following operating systems:
 
 ### 1. CentOS
 
-Install binary dependencies:
-
-```sh
-sudo yum install perl-JSON-XS
```
matsduf: When libraries are installed from binaries it will be more consistent what you get. It also helps for future packaging. I think it should be kept as-is.

docs/Installation.md
```diff
@@ -38,18 +38,6 @@ This instruction covers the following operating systems:
...
-```
-
-Install dependencies from CPAN:
-
-```sh
-sudo cpanm Locale::Msgfmt Locale::TextDomain MooseX::Getopt Text::Reflow
```
matsduf: If dependencies are explicitly installed from CPAN or indirectly installed when Zonemaster::CLI is installed from CPAN makes no difference. I cannot see the value of keeping this.

mattias-p: Maintaining the list of yum packages is significantly easier when we have this list spelled out.

matsduf: Do you mean "searching for RPMs is significantly easier when we have this list spelled out"? -- true.

mattias-p: Well, I think the greatest improvement is in knowing what RPMs to search for.

matsduf: True.

mattias-p:
@sandoche2k Can this be closed?

PR #138, opened 2020-04-30
added Dockerfile

andreasschulze:
Hi Mats,

with this pull request I like to contribute a Dockerfile to build zonemaster-cli as docker container based on Debian (latest / bullseye)

Andreas


matsduf:
Thank you. It is a great idea. I do not know why your PR fails in Travis. We have to check that. 

I have some questions and comments on your suggestion:

1. Your PR is just  a "Dockerfile". It would also need some documentation on how to use it, how to create the docker image. At least some steps, or else it will not be easier to go that direction instead of installing from the regular instructions unless you are already familiar with Docker.

2. If this should be seen as an alternative to install Zonemaster-CLI, you probably want to see that before you start installing anything. That could maybe be on the main README or in the installation document for Zonemaster-Engine.

3. We just released v2019.2 of Zonemaster, and new packages have been uploaded to CPAN. It should probably be tested against those. Maybe the packages in the Dockerfile should be fixed by version and then updated for each release?

4. You include "ldnsutils" as binary package. It is currently version 1.7.0, which means that it does not support DNSSEC algorithm 15. For that it requires version 1.7.1 and OpenSSL 1.1.1 (see information in [main README] and [Zonemaster-LDNS README]). 

5. I think it would be better with a solution that supports algorithm 15 because it important that Zonemaster can handle that i a good way. The algorithm is still not in general use, but it is one of the recommended algorithms by [RFC 8624].

[RFC 8624]: https://www.rfc-editor.org/rfc/rfc8624.html#section-3.1
[main README]: https://github.com/zonemaster/zonemaster/blob/af04e3746750f49c63c6d36739f31d60cf09c7d2/README.md#support-of-dnssec-algorithm-15-ed25519
[Zonemaster-LDNS README]: https://github.com/zonemaster/zonemaster-ldns/blob/2918d896d014a19f5e1ff28765ffd121994335de/README.md

matsduf:
@andreasschulze, this PR has been around for a while. It is a good proposal, and I think it would be good complete it as an alternative to "normal" installation. I am willing to do work on this too. I have a Mac laptop and should be able to run the Docker environment.

Please look at my bullets above and comment.

andreaso:
While I'm not the Andreas who originally created the PR I do have a few general thoughts...

1) Given that there are https://github.com/zonemaster/zonemaster/releases I agree that it would make sense if the Dockerfile pinned those specific LDNS, Engine and CLI versions, to be updated for every new Zonemaster release.

2) I very much think the provided Dockerfile should create a non-root user to run _zonemaster-cli_ as. After all, by its very nature the tool parses untrusted inputs, and is hence at least technically vulnerable to attack. Also, it's such an easy thing to do, so there is really little reason not to.

3) To lower the threshold for users even more, maybe the Zonemaster project should publish its Docker image in the [GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry)? That way the only thing needed for the user would be a `docker run ...` command in the README file. I mean, if I trust the Zonemaster project enough to install its CPAN packages I imagine I ought to trust the project's Docker image as well.

Also, feel free to take a look at my personal https://github.com/andreaso/zonemaster-image setup, in case it might provide any useful ideas.

matsduf:
@andreaso, thank you for your reply. Yes, your image give as some useful ideas for how to come forward with a Docker image.

matsduf:
Thank you for the proposal. The upcoming release will contain a Docker file, and an image will be uploaded to Docker Hub.

PR #141, opened 2020-05-02
Zonemaster-CLI: merge master into develop

matsduf:
Something has been merged into master instead of develop in Octorber last year. To correct that, merge master into develop.

When this has been merged, a new PR will be created to reintroduce the Danish translations into the develop branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ perl:
     - "5.20"
     - "5.16"
 before_install:
-    - sudo apt-get install -y libidn11-dev
     - eval $(curl https://travis-perl.github.io/init) --auto
     - local-lib
     - git clone --depth=1 --branch=develop https://github.com/zonemaster/zonemaster-ldns.git
--- a/share/da.po
+++ b/share/da.po
@@ -4,139 +4,139 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: 0.0.5\n"
-"PO-Revision-Date: 2019-07-08\n"
-"Last-Translator: haarbo@dk-hostmaster.dk\n"
+"PO-Revision-Date: 2017-12-06\n"
+"Last-Translator: mats.dufberg@iis.se\n"
 "Language-Team: Zonemaster Team\n"
 "MIME-Version: 1.0\n"
 
 msgid "Print version information and exit."
-msgstr "Udskriv version og afbryd."
+msgstr "Print version information and exit."
 
 msgid "The minimum severity level to display"
-msgstr "Det mindste alvorlighedsniveau som skal udskrives."
+msgstr "The minimum severity level to display"
 
 msgid "The locale to use for messages translation."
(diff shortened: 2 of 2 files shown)


matsduf:
> @matsduf Is it correct that you have an English version in the da.po file?

No, and that should be corrected. I think it will be easier to do in two steps. First correct, and then add a change to the develop branch.

mattias-p:
I don't mean to be adverse but I have a couple of points to make here.

First, we should strive to always keep develop in a releasable state. In this case that could be achieved by creating a new local branch off of master, cherry picking the reverted commits back onto the new branch and merging all of that back into develop in a single PR.

Second, this operation is not according to our processes. Our current process is to only merge develop into master and not the other way around. I don't see much of a point in merging master back into develop in an ad-hoc manner. That said, if we do want to merge master back into develop, we should update the release process with a merge-back-into-develop step and perform that merge on all the repos.

N.B. The merge-back-into-develop step could include an instruction to reapply any commits that were reverted on the master branch.

matsduf:
> First, we should strive to always keep develop in a releasable state. In this case that could be achieved by creating a new local branch off of master, cherry picking the reverted commits back onto the new branch and merging all of that back into develop in a single PR.

The develop branch is in releasable state. We did the release from the develop branch. This will be an update to the develop branch, and if you want to do it some other way, please create a pull request.

> Second, this operation is not according to our processes. Our current process is to only merge develop into master and not the other way around. I don't see much of a point in merging master back into develop in an ad-hoc manner. 

This would not be the first time merging master into develop. I have created such a pull request before on your proposal.

> That said, if we do want to merge master back into develop, we should update the release process with a merge-back-into-develop step and perform that merge on all the repos.

Fine for me. I have checked all repos and this is the only repo where such an operation would do any change to any file. But changing the release process is not part of this pull request.

> N.B. The merge-back-into-develop step could include an instruction to reapply any commits that were reverted on the master branch.

That comment is not for this pull request, is it?

mattias-p:
> > First, we should strive to always keep develop in a releasable state. In this case that could be achieved by creating a new local branch off of master, cherry picking the reverted commits back onto the new branch and merging all of that back into develop in a single PR.
> 
> The develop branch is in releasable state. We did the release from the develop branch. This will be an update to the develop branch, and if you want to do it some other way, please create a pull request.

Okay, I'll do that.

> > That said, if we do want to merge master back into develop, we should update the release process with a merge-back-into-develop step and perform that merge on all the repos.
> 
> Fine for me. I have checked all repos and this is the only repo where such an operation would do any change to any file. But changing the release process is not part of this pull request.

I'll create a PR for an update to the release process.

> > N.B. The merge-back-into-develop step could include an instruction to reapply any commits that were reverted on the master branch.
> 
> That comment is not for this pull request, is it?

Correct.

matsduf:
@sandoche2k and @mattias-p, how do you want to fix the issue? As it is now, the package on CPAN is different in one file from master branch.

mattias-p:
I made an alternative proposal in #142.

matsduf:
Replaced by #142.

PR #142, opened 2020-05-15
Solves inconsistency in master branch.

mattias-p:
The master branch is currently different than the latest release on CPAN. In particular the file `share/da.po` is different.

The inconsistency occurred because we accidentally merged a PR for that file into the master branch instead of the develop branch. To mend the situation we added a revert commit to the master branch to get its head back in sync with the latest release at the time. The changes to `da.po` were also merged into the develop branch in a separate PR where they were meant to go in the first place. 

Later we made a new release and merged the develop branch into the master branch as usual. Git noticed two identical changes to `da.po` on both branches and a revert commit for the change on the master branch. After merging the revert commit stays effective on the master branch. The dist tarball that was uploaded to CPAN was built from the develop branch where there is no revert commit. Thus the master branch and CPAN release are inconsistent.

We can't really remove the revert commit from the master branch, but we can revert it. Doing that brings the master branch into sync with the CPAN release.

This reverts commit 9e94ee12a50f4ad8ce61de5f40c1744ad2f8667a.

### DIFF
--- a/share/da.po
+++ b/share/da.po
@@ -4,139 +4,139 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Project-Id-Version: 0.0.5\n"
-"PO-Revision-Date: 2017-12-06\n"
-"Last-Translator: mats.dufberg@iis.se\n"
+"PO-Revision-Date: 2019-07-08\n"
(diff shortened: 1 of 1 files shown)


matsduf:
This PR replaces PR #141.

matsduf:
@mattias-p, when this is merged I will move the release to the top of master branch.

PR #150, opened 2020-06-08
Fix typo

mattias-p:
This change brings the source into sync with the PO files for this particular message.


mattias-p:
No, of course not. Good catch!

PR #151, opened 2020-06-08
Update translation workflow

### LINKED ISSUE #132: share/Makefile not compatible with FreeBSD "make"
The share/Makefile contains the GNU Make "wildcard" macro not compatible with  FreeBSD. Also see zonemaster/zonemaster-engine#697.

### LINKED ISSUE #147: Mimic translation tooling from Engine
Today the process for updating the PO files in CLI is completely manual. On the other hand Engine uses `xgettext` and `msgmerge` to do much of the heavy lifting. CLI should adopt the same tooling as Engine.

mattias-p:
This PR enables the adoption in CLI of the translation workflow we use in Engine.
This will affect the format of the PO files making for a large initial diff. That initial update is excluded from this PR because the PO files are currently not up to date.

Fixes #132. Fixes #147.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ cover_db/
 blib/
 inc/
 share/locale/
+share/*.mo
 .lwpcookies
 .last_cover_stats
 nytprof.out
@@ -18,3 +19,4 @@ pm_to_blib
 Zonemaster-CLI-*
 Zonemaster-CLI-*.tar.gz
 *.bak
+share/Zonemaster-CLI.pot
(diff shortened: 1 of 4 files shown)


matsduf:
> That initial update is excluded from this PR because the PO files are currently not up to date.

This refers to the initial update of the PO files, doesn't it?

We need translation instructions for CLI too, but since it changes less often in CLI it might be possible to just refer to the instructions for Zonemaster-Engine?

mattias-p:
>> That initial update is excluded from this PR because the PO files are currently not up to date.
>
> This refers to the initial update of the PO files, doesn't it?

Correct.

> We need translation instructions for CLI too, but since it changes less often in CLI it might be possible to just refer to the instructions for Zonemaster-Engine?

Yeah, I think duplicating them would cause more harm than good.

PR #157, opened 2020-07-23
Related to zonemaster-engine issue 578

vlevigneron:
- Add method name (test case) in Logger.
- Fixes zonemaster/zonemaster-engine#578

DO NOT MERGE NOW, do a test first to check if it is what you are waiting for

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -6,10 +6,12 @@ extends 'Zonemaster::Engine::Exception';
 # The actual interesting module.
 package Zonemaster::CLI;
 
-use version; our $VERSION = version->declare("v2.0.4");
-
 use 5.014002;
+
+use strict;
 use warnings;
+
+use version; our $VERSION = version->declare("v2.0.5");
 
 use Locale::TextDomain 'Zonemaster-CLI';
 use Moose;
@@ -109,6 +111,13 @@ has 'show_module' => (
     is            => 'ro',
     isa           => 'Bool',
     documentation => __( 'Print the name of the module on entries.' ),
+    default       => 0,
+);
+
+has 'show_testcase' => (
+    is            => 'ro',
+    isa           => 'Bool',
+    documentation => __( 'Print the name of the test case (method) on entries.' ),
     default       => 0,
 );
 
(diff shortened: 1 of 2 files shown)


matsduf:
@vlevigneron, does this PR depend on the code change by zonemaster/zonemaster-engine/pull/762?

lib/Zonemaster/CLI.pm
```diff
@@ -337,7 +346,7 @@ sub run {
         }
 
         if ( $self->config ) {
-            say $fh_diag __x( "Loading configuration from {path}.", path => $self->config );
+            say $fh_diag __x( "Loading configuaration from {path}.", path => $self->config );
```
mattias-p, line 349: `configuaration` -> `configuration`

lib/Zonemaster/CLI.pm
```diff
@@ -397,20 +406,25 @@ sub run {
                     }
 
                     if ( $self->show_level ) {
-                        printf "%-9s ", translate_severity( $entry->level );
+                        printf "%-9s ", __( $entry->level );
```
mattias-p, line 409: This breaks msgid extraction. `xgettext` depends on `__()` only ever being called with literal (format) strings.

lib/Zonemaster/CLI.pm
```diff
@@ -112,6 +114,13 @@ has 'show_module' => (
     default       => 0,
 );
 
+has 'show_testcase' => (
+    is            => 'ro',
+    isa           => 'Bool',
+    documentation => __( 'Print the name of the test case (method) on entries.' ),
```
mattias-p, line 120: I don't think we should refer to test cases as methods at the interface level. The interface documentation shouldn't leak implementation details such as how test cases are implemented in code.

mattias-p (changes requested):

matsduf:
@vlevigneron, can this be completed?

lib/Zonemaster/CLI.pm
vlevigneron: Fixed

lib/Zonemaster/CLI.pm
vlevigneron: Sorry, I missed the point. It has been fixed.

lib/Zonemaster/CLI.pm
vlevigneron: You are right, I modify that.

vlevigneron:
@matsduf @mattias-p Please review again and check if output is what you are waiting for.

matsduf (approved):

matsduf:
@mattias-p, can you review again?

mattias-p (approved):

PR #158, opened 2020-07-24
Skip MO files directly under share

matsduf:
PR #151 updated the way the MO files are generated. Now two copies are created and the one directly under /share should be skipped. PR #151 missed to include a SKIP statement in MANIFEST.SKIP, which is done here, modeled after the statement in Zonemster-Engine.


matsduf:
@vlevigneron, can you review? I think this should be straight forward.

PR #167, opened 2020-10-15
Corrected listing of files

matsduf:
This corrects the same issue as zonemaster/zonemaster-engine/pull/804 but in CLI. "-depth 1" has been added to really mirror what the original code wanted to do.

### DIFF
--- a/share/Makefile
+++ b/share/Makefile
@@ -3,7 +3,7 @@
 .PHONY: all dist touch-po update-po extract-pot
 
 DISTNAME = Zonemaster-CLI
-POFILES != echo *.po
+POFILES != find . -type f -name '*.po' -depth 1
 MOFILES := $(POFILES:%.po=%.mo)
 POTFILE = $(DISTNAME).pot
 PMFILES != find ../lib -type f -name '*.pm' | sort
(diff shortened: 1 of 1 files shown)


share/Makefile
```diff
@@ -3,7 +3,7 @@
 .PHONY: all dist touch-po update-po extract-pot
 
 DISTNAME = Zonemaster-CLI
-POFILES != echo *.po
+POFILES != find . -type f -name '*.po' -depth 1
```
mattias-p, line 6: I think you meant `-maxdepth 1`

matsduf: No, I meant `-depth 1`, but when the value is 1 both give the same result. If you see any problem with `-depth` that `-maxdepth` does not have I can change.

And as you commented in zonemaster/zonemaster-engine/pull/804, we do not expect any other PO files in that subtree so with and without `-(max)depth 1` will give that same result in practice.

matsduf: I see that `-depth n` does not seem to be in Linux find(). Changed to `-maxdepth 1`. And it also wants that to be before the file pattern. See update.

matsduf:
@mattias-p, @pnax and @sandoche2k, please review.

mattias-p (approved):

PR #169, opened 2020-10-24
Fix make and makefile

matsduf:
CLI has the same problem as Engine with `share/Makefile` not being compatible with GNU Make on CentOS 7 (zonemaster/zonemaster-engine#813). The solution, as implemented by zonemaster/zonemaster-engine#814, made `share/Makefile` incompatible with BSD Make (on FreeBSD).

A partial solution to the problem for FreeBSD is proposed in zonemaster/zonemaster-engine#815.

This PR proposes a complete solution to the problem. GNU Make and BSD Make cannot sucessfully use the same Makefile, but GNU Make can be made available on FreeBSD as `gmake`. And for GNU Make the file `GNUmakefile` is the preferred file, if available. We use that fact.

* Update `Makefile` to be GNU Make compatible for CentOS 7 (and incompatible for BSD Make).
* Rename `Makefile` to `GNUmakefile`.
* Create wrapper `Makefile` for BSD Make to make it run the same target as `gmake` (and `GNUmakefile`).
* Update `Makefile.PL` to require `gmake` for FreeBSD and to use `gmake` for `share/GNUmakefile` under FreeBSD.
* Update installation instructions to include the installation of `gmake`.

The separation into two makefiles will make it possible to update those for future use.


matsduf:
@sandoche2k, before you test build CLI on CentOS 7, review and test this PR.

Makefile.PL
```diff
@@ -23,13 +23,30 @@ requires(
...
-	$(MYEXTLIB): share/Makefile
-	cd share && $(MAKE) all
-MAKE_FRAG
+    my $sharemakefile = 'share/GNUmakefile';
+    my $text;
+    if ($^O eq "freebsd") {
+        # Make FreeBSD use gmake for share
+        $text = q[GMAKE ?= "gmake"] . "\n"
+            . q[pure_all :: ] . "$sharemakefile\n"
+            . "\t" . q[cd share && $(GMAKE) all] . "\n";
```
mattias-p, line 39: Could you make the string quoting consistent? At the very least locally, but preferably also with the rest of the repo. E.g. `q[]` isn't used anywhere else in zonemaster-engine. Also, mixing concatenation and interpolation only makes it harder to read.

Makefile.PL
```diff
@@ -23,13 +23,30 @@ requires(
...
+    } else {
+        # Here Linux and GNU Make is assumed
+        $text = q[pure_all :: ] . "$sharemakefile\n"
+            . "\t" . q[cd share && $(MAKE) all] . "\n";
+    };
```
mattias-p, line 44: Why not make both cases use the same variable for the make command?

mattias-p (changes requested):
This is a quite non-trivial change. I don't think we should be merging this kind of thing this close to the release. It's better to adopt the same solution we used in Engine. We can do loftier things for the next release.

Makefile.PL
matsduf: The $(MAKE) variable comes from the Makefile. The variable is interpolated by make.

Makefile.PL
matsduf: Updated, but no change of logic.

matsduf:
@mattias-p and @sandoche2k, please review again.

Makefile.PL
matsduf: The changed has been tested for both FreeBSD and Ubuntu, and passed.

sandoche2k (approved):

Makefile.PL
mattias-p: I think I would have gone for setting GMAKE to different values on different platforms. That way the "pure_all" clause could be the same for all platforms.

mattias-p:
@matsduf I can't point to any technical errors in this PR, but do you really think we should make changes of this complexity this close to the release? We have much greater chances of finding problems with the Makefile-GNUmakefile approach if we merge it in the beginning of a release cycle. I think we should go for a more modest solution for this release.

matsduf:
> We have much greater chances of finding problems with the Makefile-GNUmakefile approach if we merge it in the beginning of a release cycle.

It is well-documented that GNU Make prefers `GNUmakefile` before `Makefile`. The only problem would be the wrapper function, but that only affects FreeBSD which does not work well with the current situation anyway. I do not see that the risks are higher to implement than to keep it.

I am surprised that we have not seen the solution before instead of trying to make `Makefile` work for both GNU Make and BSD Make.

Makefile.PL
matsduf: Such as "GMAKE = $(MAKE)" and "GMAKE = gmake", respectively?

Well, it has the same effect. Please the updated code.

mattias-p:
It's true that GNU make will pick the GNUmakefile.

But it's no beauty queen. I guess that could be a reason why we haven't seen it before. For the next release I'll see if I can't come up with something more satisfying.

mattias-p (approved):

Makefile.PL
mattias-p: Yeah, that's what I had in mind. I'm thinking that it's a little easier on the reader (i.e. us) if you don't have to compare two heavily quoted and generated makefile snippets by eye.

matsduf:
Unless the makefile is very simple it does not seem to be possible to share it between BSD Make and GNU Make. I think this solution of having a wrapper for BSD Make is quite sound. Else we could leave `make` and use a shell script instead.

PR #171, opened 2020-10-28
Fix po files

pnax:
Add missing `.` in some msgid and update the language value in sv file.

### DIFF
--- a/share/da.po
+++ b/share/da.po
@@ -12,7 +12,7 @@ msgstr ""
 msgid "Print version information and exit."
 msgstr "Udskriv version og afbryd."
 
-msgid "The minimum severity level to display"
+msgid "The minimum severity level to display."
 msgstr "Det mindste alvorlighedsniveau som skal udskrives."
 
 msgid "The locale to use for messages translation."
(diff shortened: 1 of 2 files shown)


share/da.po
```diff
@@ -12,7 +12,7 @@ msgstr ""
 msgid "Print version information and exit."
 msgstr "Udskriv version og afbryd."
 
-msgid "The minimum severity level to display"
+msgid "The minimum severity level to display."
```
matsduf, line 15: That is not correct msgid from the code.

share/sv.po
```diff
@@ -12,7 +12,7 @@ msgstr ""
 msgid "Print version information and exit."
 msgstr "Skriv ut versionsinformation och avsluta."
 
-msgid "The minimum severity level to display"
+msgid "The minimum severity level to display."
```
matsduf, line 15: This is not correct msgid from the code.

matsduf (changes requested):
The msgid should be updated by first doing any updates of the Perl module and then generate new PO files by e.g. "make da.po". By doing that, msgids will be correct.

mattias-p:
It's important that the msgid is identical to the string given to the `__(...)` function (or one of its friends). That's one of the core principle of how the gettext system works.

But it looks to me like we've forgotten to ask for new translations for CLI. I'll make a PR that fixes the formatting of the PO files to make life easier for our translators.

share/da.po
pnax: Yes indeed, I thought the rest of the msgid was wrapped on a second line.

share/sv.po
pnax: Yes, my mistake. And thanks for the tip with the make command ;)

pnax:
> It's important that the msgid is identical to the string given to the __(...) function (or one of its friends). That's one of the core principle of how the gettext system works.

Yes, I thought this was just a missing dot. I'll be more careful with my greps.

I am closing this PR because this is not the proper way of updating po files, there is a make command for that.

Issue #182, opened 2020-12-10
Selected profile overrides IPv6 setting 


matsduf:
If both profile (`--profile`) and IPv6 (`--no-ipv6`/--ipv6`) are selected then the IPv6 setting will be overridden by the profile, but the other way around is expected.

First run the test with `--no-ipv6` but no profile selected. Note the messages "IPv6 is disabled".
```
$ zonemaster-cli --no-ipv6 zonemaster.net --locale en_US.UTF-8 --test basic --level INFO
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.0.3 of the Zonemaster engine.
   1.06 INFO      Parent domain 'net' was found for the tested domain.
   4.60 INFO      IPv6 is disabled, not sending "NS" query to ns.nic.se/2001:67c:124c:100a::45.
   4.60 INFO      IPv4 is enabled, can send "NS" query to ns.nic.se/91.226.36.45.
   4.62 INFO      Nameserver ns.nic.se/91.226.36.45 listed these servers as glue: ns.nic.se.,ns2.nic.fr.,ns3.nic.se..
   4.62 INFO      IPv4 is enabled, can send "NS" query to ns2.nic.fr/192.93.0.4.
   4.67 INFO      Nameserver ns2.nic.fr/192.93.0.4 listed these servers as glue: ns.nic.se.,ns2.nic.fr.,ns3.nic.se..
   4.67 INFO      IPv6 is disabled, not sending "NS" query to ns2.nic.fr/2001:660:3005:1::1:2.
   4.67 INFO      IPv6 is disabled, not sending "NS" query to ns3.nic.se/2001:67c:124c:2007::45.
   4.68 INFO      IPv4 is enabled, can send "NS" query to ns3.nic.se/91.226.37.45.
   4.69 INFO      Nameserver ns3.nic.se/91.226.37.45 listed these servers as glue: ns.nic.se.,ns2.nic.fr.,ns3.nic.se..
   4.69 INFO      Functional nameserver found. "A" query for www.zonemaster.net test skipped.
```

Again run with `--no-ipv6` but also select a profile, which in this case is the same profile as would be used anyway. Note the messages "IPv6 is enabled".
```
$ zonemaster-cli --no-ipv6 zonemaster.net --locale en_US.UTF-8 --test basic --level INFO --profile /usr/local/lib/perl5/site_perl/auto/share/dist/Zonemaster-Engine/profile.json
[...]

PR #185, opened 2021-01-19
Updates the ipv4 and ipv6 command line options

matsduf:
Makes the ipv4 and ipv6 options undefined unless explicitly set on the command line. Changes so that the ip setting is only changed if the option is explicitly set on the command line.

This resolves a bug that appeared after #182 was fixed by #183.

The bug is described in  https://github.com/zonemaster/zonemaster-cli/issues/182#issuecomment-762787971.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -357,10 +357,10 @@ sub run {
 
     # These two must come after any profile from command line has been loaded
     # to make any IPv4/IPv6 option override the profile setting.
-    if defined ($self->ipv4) {
+    if ( defined ($self->ipv4) ) {
         Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
     }
-    if defined ($self->ipv6) {
+    if ( defined ($self->ipv6) ) {
         Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
     }
 
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -147,7 +147,6 @@ has 'restore' => (
 has 'ipv4' => (
     is      => 'ro',
     isa     => 'Bool',
-    default => 1,
     documentation =>
       __( 'Flag to permit or deny queries being sent via IPv4. --ipv4 permits IPv4 traffic, --no-ipv4 forbids it.' ),
 );
@@ -155,7 +154,6 @@ has 'ipv4' => (
 has 'ipv6' => (
     is      => 'ro',
     isa     => 'Bool',
-    default => 1,
     documentation =>
       __( 'Flag to permit or deny queries being sent via IPv6. --ipv6 permits IPv6 traffic, --no-ipv6 forbids it.' ),
 );
@@ -358,9 +356,9 @@ sub run {
     }
 
     # These two must come after any profile from command line has been loaded
-    # to override the profile setting
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    # to override the profile setting, but only if selected.
+    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 ) if defined ($self->ipv4);
+    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 ) if defined ($self->ipv6);
 
 
     if ( $self->dump_profile ) {
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -362,7 +362,7 @@ sub run {
     }
     if defined ($self->ipv6) {
         Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
-    };
+    }
 
 
     if ( $self->dump_profile ) {
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -356,9 +356,13 @@ sub run {
     }
 
     # These two must come after any profile from command line has been loaded
-    # to override the profile setting, but only if selected.
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 ) if defined ($self->ipv4);
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 ) if defined ($self->ipv6);
+    # to make any IPv4/IPv6 option override the profile setting.
+    if defined ($self->ipv4) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
+    }
+    if defined ($self->ipv6) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    };
 
 
     if ( $self->dump_profile ) {


lib/Zonemaster/CLI.pm
```diff
@@ -358,9 +356,9 @@ sub run {
     }
 
     # These two must come after any profile from command line has been loaded
-    # to override the profile setting
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    # to override the profile setting, but only if selected.
+    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 ) if defined ($self->ipv4);
+    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 ) if defined ($self->ipv6);
```
mattias-p, line 361: Could you update this to use `if (...) { ... }` syntax? This is in line with the rest of the code. Also, the structure of the code is more apparent to the reader that way.

I do realize there are a couple of instances of trailing if statements in this file, but adding more of them is a mistake from a readability point of view.

lib/Zonemaster/CLI.pm
```diff
@@ -358,9 +356,9 @@ sub run {
     }
 
     # These two must come after any profile from command line has been loaded
-    # to override the profile setting
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    # to override the profile setting, but only if selected.
```
mattias-p, line 359: If the control flow is apparent from the syntax (like I suggest below), this added clarification is unnecessary and distracting. And it's better to remove it.

lib/Zonemaster/CLI.pm
matsduf: Is that only your personal opinion that trailing "if" is always less readable, or do you know of any studies of that supports that? In that case we get three lines, instead of one, and more separators. And the condition is short.

mattias-p: Are you being serious?

https://www.squirrel.nl/pub/PBP_refguide-1.02.00.pdf

lib/Zonemaster/CLI.pm
```diff
@@ -147,15 +147,13 @@ has 'restore' => (
 has 'ipv4' => (
     is      => 'ro',
     isa     => 'Bool',
-    default => 1,
```
pnax, line 150: This default behavior is not moved into the code. I mean, before the change if you do not have the `net` stanza in the profile file and you just run the CLI, IPv4 and IPv6 will be enabled by default.

Here with the current changes, removing the stanza from the profile file gives a CRITICAL message :
```
   0.01 CRITICAL  Both IPv4 and IPv6 are disabled.
```

I feel an `else` statement is needed in the code to enable IPv4 and IPv6 by default.

lib/Zonemaster/CLI.pm
```diff
@@ -358,9 +356,9 @@ sub run {
     }
 
     # These two must come after any profile from command line has been loaded
-    # to override the profile setting
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    # to override the profile setting, but only if selected.
+    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 ) if defined ($self->ipv4);
```
pnax, line 360: The `else` statement I see :

```
else {
    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 1 );
}
```

lib/Zonemaster/CLI.pm
matsduf: I had to remove that "default" to make it possible to detect if the option had been actively set or not.

lib/Zonemaster/CLI.pm
matsduf: That "else" should first check if any of ipv4/ipv6 is enabled, if not set both. Otherwise we will, again, override the setting in any selected config file.

I will check and come with an updated version.

pnax: Oh yes I didn't see that it was as twisted as that.

I think this cumbersome code covers every cases (maybe some perl magic might make it lighter)

```
    if ( defined ($self->ipv4) ) {
        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, $self->ipv4 );
    } else {
        if ( ! defined ($self->ipv6) ) {
            Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 1 );
        }
    }

    if ( defined ($self->ipv6) ) {
        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, $self->ipv6 );
    } else {
        if ( ! defined ($self->ipv4) ) {
            Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 1 );
        }
    }
```

matsduf: This will also be wrong, because if you have selected a profile where IPv6 has been disabled, then your suggestion will enable IPv6 unless you explicitly have selected --no-ipv6.

I will proved an update.

pnax: This is really twisted ^^. I'll let you find a solution ;)

lib/Zonemaster/CLI.pm
matsduf: Now I see what you mean. If the default profile.json does not have any IPv4 or IPv6 enabled, then there will no network enabled. -- In Backend both are enabled if both are disabled.

matsduf:
@mattias-p and @pnax, should CLI be "helpful" and enable both IPv4 and IPv6 if both are disabled? I do not think so. I think it is better that the user gets the message rather than adding that magic.

mattias-p:
I don't think we should support modifying the default profile.

If a user specifies a profile with both IPv4 and IPv6 disabled but forgets to override either one on the command line, it seems reasonable to just let them know that nothing can be done because both IPv4 and IPv6 are disabled. Maybe it would be better to give the user feedback in the form of an input validation error than in the form of a message tag.

lib/Zonemaster/CLI.pm
```diff
@@ -356,9 +356,13 @@ sub run {
...
-    # to override the profile setting, but only if selected.
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 ) if defined ($self->ipv4);
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 ) if defined ($self->ipv6);
+    # to make any IPv4/IPv6 option override the profile setting.
+    if defined ($self->ipv4) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
+    }
+    if defined ($self->ipv6) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    };
```
pnax, line 365: unnecessary trailing semicolon

And I think that the `0+` can be removed as well.

matsduf:
> I don't think we should support modifying the default profile.

I agree.

> Maybe it would be better to give the user feedback in the form of an input validation error than in the form of a message tag.

That would be beyond the scope of this PR for fixing the bug late in the release cycle. I suggest a issue to do that update in next release.

pnax (commented):
From my point of view, if IPv4 and/or IPv6 are **explicitly** disabled, then the CLI should stake to this choice.

However, I think this does not solve a specific case. From my understanding the `net` stanza in the profile is not mandatory. Hence if the profile.json file do not have the `net` stanza (because the user forgot to add it, because it was shipped without it...) then there is no default fallback. I think the CLI should use IPv4 and IPv6 by default in this specific case.
Or maybe this stanza is already mandatory and I missed this.

Is there any documentation about the precedence between the CLI `--(no)ipvx` options and the profile.json entries ?

lib/Zonemaster/CLI.pm
mattias-p: > And I think that the 0+ can be removed as well.

That's correct. It's in the contract of the Profile properties to always return the correct data type.

matsduf:
> From my point of view, if IPv4 and/or IPv6 are **explicitly** disabled, then the CLI should comply with this choice.

If you select a profile with "--profile" and that has disabled both or you have  disable by "--no-ipv4 --no-ipv6" then I think we can say that the user has **explicitly** disbled both.

> However, I think this does not solve a specific case. From my understanding the `net` stanza in the profile is not mandatory. Hence if the profile.json file do not have the `net` stanza (because the user forgot to add it, because it was shipped without it...) then there is no default fallback. I think the CLI should use IPv4 and IPv6 by default in this specific case.

The `profile.json` is part of the installation, and that is **always** loaded. We should not support editing of that. That profile has both ipv4 and ipv6 enabled.

If you choose another profile by "--profile" then first the default `profile.json` will be loaded and then your profile, which will override a parameters that are defined in your profile.

> Or maybe this stanza is already mandatory and I missed this.

It is not mandatory, but is is included in the default `profile.json` that is provided by the installation and **always** loaded.

> Is there any documentation about the precedence between the CLI `--(no)ipvx` options and the profile.json entries ?

No, not explicitly.

lib/Zonemaster/CLI.pm
matsduf: Concerning "0+" I also looked at that, but I wanted to change as little as possible since this is a bug fix late in the cycle.

The tailing semicolon it not needed. When I started with Perl 1996 I think it was recommended, or maybe even mandatory in Perl 4, so I have a habit of writing them. It does not harm anything.

mattias-p:
@pnax I feel there's a misunderstanding here.

The Profile module was designed to never have any undefined properties in the default profile, nor in the effective profile. They always start out with a valid default value - in particular the default value that is specified in the documentation (i.e. true for both net.ipv4 and net.ipv6). If users modify the default profile, they're breaking the Profile contract and all bets are off.

Users should be able to override the default profile with --profile and other arguments. So the only way to get to no-IPv4 and no-IPv6 should be to specify that on the command line.

Lastly, no, I don't think we have any documentation on precedence. We should have that.

lib/Zonemaster/CLI.pm
mattias-p: I appreciate the historical note :)

pnax:
Indeed this is not the logic I had in mind. Thank you for your insights.

matsduf:
@mattias-p and @pnax, is this good enough? Can you approve?

lib/Zonemaster/CLI.pm
```diff
@@ -358,9 +356,13 @@ sub run {
...
-    # to override the profile setting
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
-    Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    # to make any IPv4/IPv6 option override the profile setting.
+    if defined ($self->ipv4) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv4}, 0+$self->ipv4 );
+    }
+    if defined ($self->ipv6) {
+        Zonemaster::Engine::Profile->effective->set( q{net.ipv6}, 0+$self->ipv6 );
+    }
```
pnax, line 365: I tested it, and it looks like the parentheses usage in the if statement makes perl comply.

mattias-p:
@matsduf Yeah, just fix the syntax errors and I'll approve.

lib/Zonemaster/CLI.pm
matsduf: I should have kept the first version... Fix in a minute.

matsduf:
I have verified that the code works after my last commit. @mattias-p and @pnax, please review.

pnax (approved):

PR #186, opened 2021-01-20
Avoid updating PO files for clean repo

mattias-p:
When I run `perl Makefile.PL` on a fresh repository in preparation for running `make dist`, the PO files are updated with fuzzy translations. Engine does not have this problem. This PR adopts the solution from Engine, making sure that the PO files are not updated when running Makefile.PL.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -39,11 +39,11 @@ sub MY::postamble {
         # Make FreeBSD use gmake for share
         return "GMAKE ?= \"gmake\"\n"
             . "pure_all :: $sharemakefile\n"
-            . "\tcd share && \$(GMAKE) all\n";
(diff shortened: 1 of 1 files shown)


matsduf:
@mattias-p, this is a bug, isn't it? And should be merged for v2020.2, shouldn't it?

matsduf (approved):

mattias-p:
It's a bug in the build system. And I believe it's important that it's included in the release. Otherwise I believe we run the risk of installing fuzzy messages on end-user systems. I forgot to tag the PR, sorry about that.

matsduf:
> I forgot to tag the PR, sorry about that.

Don't worry. I've taken that to be my task to check. :-)

PR #188, opened 2021-01-20
Update dependencies installation on CentOS

mattias-p:
* Replace MooseX::Getopt with perl-MooseX-Getopt
* Remove Locale::MsgFmt (not used)
* Remove Locale::TextDomain (use the same one as Engine)
* Work around compolation errors on CentOS 8

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -46,8 +46,18 @@ This instruction covers the following operating systems:
 2) Install dependencies from CPAN:
 
    ```sh
-   sudo cpanm Net::Interface Text::Reflow
+   sudo cpanm Text::Reflow
+   curl -O https://cpan.metacpan.org/authors/id/M/MI/MIKER/Net-Interface-1.016.tar.gz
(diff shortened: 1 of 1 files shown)


matsduf (changes requested):
Since the work-around is only needed for CentOS 8 I think that CentOS 7 should have a normal installation.

mattias-p:
Are you sure? In my opinion it's more important that the installation instructions are easy to follow and maintain than that they result in the perfect installation. I'd prefer if users could just breeze through the installation as much as possible and stop and think as little as possible. I think we should prioritize onboarding.

matsduf:
> Are you sure? In my opinion it's more important that the installation instructions are easy to follow and maintain than that they result in the perfect installation. I'd prefer if users could just breeze through the installation as much as possible and stop and think as little as possible. I think we should prioritize onboarding.

That is not how we have done for Zonemaster-Engine and Zonemaster-Backend. I think that has worked well.

matsduf (approved):

pnax (approved):

PR #208, opened 2021-05-31
Fixes error that prevents MO file to be created

matsduf:


mattias-p (approved):

mattias-p:
I was a able to reproduce this with the following commands:
```
cd share
rm nb.mo
make nb.mo
```
To which I got this output:
```
nb.po:175: 'msgid' and 'msgstr' entries do not both end with '\n'
msgfmt: found 1 fatal error
make: *** [GNUmakefile:40: nb.mo] Error 1
```
Adding the newline allowed me to create the MO file.

matsduf:
I think we should fix the msgid in CLI so that they do not end with neither newline nor space.

Issue #159, opened 2020-07-29
Make options more user friendly


matsduf:
Multi-word options, like "show module", uses underscore "_" instead of hyphen "-" as a substitution for space. I.e.:

```
--show_module
--no-show_module
```

It would be more user-friendly with 

```
--show-module
--no-show-module
```

Since underscore has been used for a long time, I suggest that both are supported.

matsduf:
@vlevigneron, would this be simple or is it a big change?

matsduf:
Resolved by #212.

PR #212, opened 2021-07-29
Make options more user friendly

vlevigneron:
## Purpose

This PR add option names with "-" (hyphen) for options with "_" (underscore) in name. There will be a backward compatibility in the next release but I suggest to remove names with underscores in the future.

## Context

Fixes zonemaster/zonemaster-cli#159

## Changes

"New" options are availables.

## How to test this PR

zonemaster-cli --list-test

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -11,7 +11,7 @@ use 5.014002;
 use strict;
 use warnings;
 
-use version; our $VERSION = version->declare( "v3.1.0" );
+use version; our $VERSION = version->declare( "v3.2.0" );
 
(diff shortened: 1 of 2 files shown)


matsduf:
> This PR add option names with "-" (hyphen) for options with "_" (underscore) in name. There will be a backward compatibility in the next release but I suggest to remove names with underscores in the future.

What problem do you of supporting both versions forever? There could be several implementations that depend on the options names with underscore.

If the plan is to remove the options names with underscore those in a near future then it should be documented in the man page that the format with underscore is deprecated.

matsduf (approved):
Works as expected.

vlevigneron:
@matsduf I see no problem. I just feel it is cleaner when --help is used for instance.

Issue #216, opened 2021-10-12
Add a Dockerfile


mattias-p:
We should provide a Docker containerization of Zonemaster CLI. This would make it easier for the end user to get a working installation of Zonemaster CLI.

matsduf:
It also opens for running zonemaster-cli on Windows or Mac.

PR #218, opened 2021-10-27
Add Dockerfile to CLI

mattias-p:
## Purpose

This PR provides dockerization of Zonemaster Engine.

## Context

This PR fixes zonemaster/zonemaster-cli#216.
Depends on zonemaster/zonemaster-engine#991.
Shout-outs to @andreasschulze, @andreaso and @blacksponge for prior work that I've built on.

## Changes

This PR adds a docker file and steps to build it.

## How to test this PR

Follow the new steps in the README file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM zonemaster/engine:local as build
+
+ARG version
+
+RUN apk add --no-cache \
+    # Only needed for Net::Interface
+    build-base \
+    perl-dev \
+    perl-doc \
+    make \
+    # Compile-time dependencies
+    perl-app-cpanminus \
+    perl-module-install \
+ && cpanm --no-wget \
+    MooseX::Getopt \
+    Text::Reflow
+
+COPY ./Zonemaster-CLI-${version}.tar.gz ./Zonemaster-CLI-${version}.tar.gz
+
+RUN cpanm --no-wget \
+    ./Zonemaster-CLI-${version}.tar.gz
+
+FROM zonemaster/engine:local
+
+COPY --from=build /usr/local/bin/zonemaster-cli /usr/local/bin/zonemaster-cli
+# Include all the Perl modules we built
+COPY --from=build /usr/local/lib/perl5/site_perl /usr/local/lib/perl5/site_perl
+COPY --from=build /usr/local/share/perl5/site_perl /usr/local/share/perl5/site_perl
+
+USER nobody:nogroup
+
+ENTRYPOINT [ "zonemaster-cli" ]
+
+CMD [ "--help" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 FROM zonemaster/engine:local as build
 
-ARG version
-
 RUN apk add --no-cache \
     # Only needed for Net::Interface
     build-base \
@@ -14,6 +12,8 @@ RUN apk add --no-cache \
  && cpanm --no-wget \
     MooseX::Getopt \
     Text::Reflow
+
+ARG version
 
 COPY ./Zonemaster-CLI-${version}.tar.gz ./Zonemaster-CLI-${version}.tar.gz
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,12 @@ RUN apk add --no-cache \
     build-base \
     perl-dev \
     perl-doc \
+    perl-lwp-protocol-https \
     make \
     # Compile-time dependencies
     perl-app-cpanminus \
     perl-module-install \
- && cpanm --no-wget \
+ && cpanm --no-wget --from https://cpan.metacpan.org/ \
     MooseX::Getopt \
     Text::Reflow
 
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    build-base \
+    perl-dev \
+    perl-doc \
+    make \
+    # Compile-time dependencies
+    perl-app-cpanminus \
+    perl-module-install \
+ && cpanm --no-wget \
+    MooseX::Getopt \
+    Text::Reflow
```
andreaso, line 16: Calling _cpanm_ like that, won't that download the packages using **http://** rather than the presumably preferred **https://**?

Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM zonemaster/engine:local as build
+
+ARG version
```
andreaso, line 3: In regards to layer caching, wouldn't it make slightly more sense to put that argument after the _apt add_ and the _cpanm_?

Dockerfile
mattias-p: You're absolutely right! I wasn't aware of this. Thanks for pointing it out!

Dockerfile
mattias-p: I thought the ARG didn't affect the caching until it was used. I'll take a closer look.

Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+COPY --from=build /usr/local/bin/zonemaster-cli /usr/local/bin/zonemaster-cli
+# Include all the Perl modules we built
+COPY --from=build /usr/local/lib/perl5/site_perl /usr/local/lib/perl5/site_perl
+COPY --from=build /usr/local/share/perl5/site_perl /usr/local/share/perl5/site_perl
+
+USER nobody:nogroup
+
+ENTRYPOINT [ "zonemaster-cli" ]
+
+CMD [ "--help" ]
```
matsduf, line 35: Translation does not work here. Can that be added?

mattias-p: In what way does it not work?

This works for me:
```sh
sudo docker run -e LANG=sv_SE zonemaster/cli:local zonemaster.net
```

README.md
```diff
@@ -27,6 +27,36 @@ Installation instructions for the CLI
...
+
+```sh
+make docker-tag-latest
+```
+
+Test a zone using the local base image:
+
+```sh
+sudo docker run zonemaster/cli:local zonemaster.net
+```
```
matsduf, line 58: To remove the container after it has run and stopped:
```sh
sudo docker run --rm zonemaster/cli:local zonemaster.net
```

mattias-p: How do you detect it's still running? When I run this I get an empty listing:

```sh
sudo docker run zonemaster/cli:local zonemaster.net ; sudo docker ps
```

matsduf:
```
sudo docker ps -a
```
lists stopped containers.

Dockerfile
matsduf: I do not get any translation with that command.

blacksponge: I tried the following commands and I can the translation in both cases
```
docker run --rm -it zonemaster/cli:local --locale fr zonemaster.net
docker run -e LANG=fr_FR.UTF-8 --rm -it zonemaster/cli:local zonemaster.net
```

Makefile.PL
```diff
@@ -33,17 +33,33 @@ if ($^O eq "freebsd") {
...
     } else {
         # Here Linux and GNU Make is assumed
-        return "pure_all :: $sharemakefile\n"
+        $pure_all = "pure_all :: $sharemakefile\n"
             . "\tcd share && \$(MAKE) touch-po all\n";
     };
+
+    my $docker = <<'END_DOCKER';
+
+docker-build: dist
```
matsduf, line 51: I guess it should be
```
docker-build: all dist
```
Unless `make all` is run before `make docker-dist` translation will not work.

Dockerfile
matsduf: It was not documented, but `make all` is required before the image is built. See https://github.com/zonemaster/zonemaster-cli/pull/218#pullrequestreview-798942149

README.md
mattias-p: Fixed

Dockerfile
mattias-p: I updated the README to include the make all target.

README.md
```diff
@@ -27,6 +27,36 @@ Installation instructions for the CLI
 
 This repository does not need any specific configuration.
 
+
+### Docker
```
blacksponge, line 31: I am not sure if the build instructions should be in the Readme. Only `docker run --rm zonemtaster/cli zonemaster.net` is relevant for someone who wants to use it

README.md
```diff
@@ -27,6 +27,36 @@ Installation instructions for the CLI
...
+Tag the local base image as the latest version:
+
+```sh
+make docker-tag-latest
+```
+
+Test a zone using the local base image:
+
+```sh
+docker run --rm zonemtaster/cli:local zonemaster.net
```
blacksponge, line 57: We could add `-t` if we want the spinner back.

README.md
matsduf: I agree. I will propose updates in a new PR when this has been merged.

matsduf (approved):

README.md
matsduf: I will include that in examples.

PR #233, opened 2021-12-01
add package install

blacksponge:
## Purpose

Provide package installation

## How to test this PR

Install zonemaster cli on debian 11 using (for the develop branch, the instruction in the doc currently install the master branch)
```sh
curl -Ls https://package.zonemaster.net/setup.sh | sudo nightly=yes bash
apt install zonemaster-cli
```

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -4,7 +4,6 @@
 
 * [Overview](#Overview)
 * [Docker](#Docker)
-* [Pre-built packages](#Pre-built-packages)
 * [Prerequisites](#Prerequisites)
 * [Local installation](#Local-installation)
   * [Installation on Rocky Linux](#Installation-on-Rocky-Linux)
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -4,7 +4,7 @@
 
 * [Overview](#Overview)
 * [Docker](#Docker)
-* [Prerequisites](#Prerequisites)
+* [Prerequisites for CPAN installation](#Prerequisites-for-CPAN-installation)
 * [Local installation](#Local-installation)
   * [Installation on Rocky Linux](#Installation-on-Rocky-Linux)
   * [Installation on Debian or Ubuntu](#Installation-on-Debian-or-Ubuntu)
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -79,9 +79,10 @@ Zonemaster::CLI, see the [declaration of prerequisites].
 Using pre-built packages is the preferred method for Debian. If you prefer to
 install from CPAN instead, follow the steps for Ubuntu.
 
-1) Configure Zonemaster packages repository
+1) Add Zonemaster packages repository to repository list
    ```sh
-   curl -Ls https://package.zonemaster.net/setup.sh | sudo sh
+   curl -Ls https://package.zonemaster.net/setup.sh
+   sudo sh setup.sh
    ```
 2) Install Zonemaster CLI
    ```sh
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -40,7 +40,7 @@ Pre-built packages are available for Debian and Ubuntu (architecture amd64 only)
 
 1) Configure Zonemaster repository
    ```sh
-   curl -Ls https://package.zonemaster.net/get.sh | sudo bash
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo bash
    ```
 2) Install Zonemaster CLI
    ```sh
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -7,7 +7,8 @@
 * [Prerequisites for CPAN installation](#Prerequisites-for-CPAN-installation)
 * [Local installation](#Local-installation)
   * [Installation on Rocky Linux](#Installation-on-Rocky-Linux)
-  * [Installation on Debian or Ubuntu](#Installation-on-Debian-or-Ubuntu)
+  * [Installation on Debian](#Installation-on-Debian)
+  * [Installation on Ubuntu](#Installation-on-Ubuntu)
   * [Installation on FreeBSD](#Installation-on-FreeBSD)
 * [Post-installation sanity check](#Post-installation-sanity-check)
 * [Using zonemaster-cli](#Using-zonemaster-cli)
@@ -33,8 +34,9 @@ not relevant for running Zonemaster-CLI on Docker.
 
 ## Prerequisites for CPAN installation
 
-Before installing Zonemaster::CLI, you should [install Zonemaster::Engine][
-Zonemaster::Engine installation].
+Before installing Zonemaster::CLI from CPAN, you should [install
+Zonemaster::Engine][ Zonemaster::Engine installation], unless you are
+to install on Debian using pre-built packages (see below).
 
 > **Note:** [Zonemaster::Engine] and [Zonemaster::LDNS] are dependencies of
 > Zonemaster::CLI. Zonemaster::LDNS has a special installation requirement,
@@ -72,8 +74,20 @@ Zonemaster::CLI, see the [declaration of prerequisites].
    ```
 
 
-### Installation on Debian or Ubuntu
-1) Update configuration of "locale"
+### Installation on Debian
+
+Using pre-built packages is the preferred method for Debian. If you prefer to
+install from CPAN instead, follow the steps for Ubuntu.
+
+1) Configure Zonemaster repository
+   ```sh
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo sh
+   ```
+2) Install Zonemaster CLI
+   ```sh
+   sudo apt install zonemaster-cli
+   ```
+3) Update configuration of "locale"
 
    ```sh
    sudo perl -pi -e 's/^# (da_DK\.UTF-8.*|en_US\.UTF-8.*|fi_FI\.UTF-8.*|fr_FR\.UTF-8.*|nb_NO\.UTF-8.*|sv_SE\.UTF-8.*)/$1/' /etc/locale.gen
@@ -90,29 +104,34 @@ Zonemaster::CLI, see the [declaration of prerequisites].
    sv_SE.utf8
    ```
 
-#### Pre-built packages (Debian only)
+### Installation on Ubuntu
 
-2) Configure Zonemaster repository
-   ```sh
-   curl -Ls https://package.zonemaster.net/setup.sh | sudo sh
-   ```
-3) Install Zonemaster CLI
-   ```sh
-   sudo apt install zonemaster-cli
-   ```
-
-#### CPAN Installation
-
-2) Install dependencies:
+1) Install dependencies:
 
    ```sh
-   sudo apt-get install libmoosex-getopt-perl libtext-reflow-perl libmodule-install-perl
+   sudo apt-get install locales libmoosex-getopt-perl libtext-reflow-perl libmodule-install-perl
    ```
 
-3) Install Zonemaster::CLI:
+2) Install Zonemaster::CLI:
 
    ```sh
    sudo cpanm Zonemaster::CLI
+   ```
+3) Update configuration of "locale"
+
+   ```sh
+   sudo perl -pi -e 's/^# (da_DK\.UTF-8.*|en_US\.UTF-8.*|fi_FI\.UTF-8.*|fr_FR\.UTF-8.*|nb_NO\.UTF-8.*|sv_SE\.UTF-8.*)/$1/' /etc/locale.gen
+   sudo locale-gen
+   ```
+
+   After the update, `locale -a` should at least list the following locales:
+   ```
+   da_DK.utf8
+   en_US.utf8
+   fi_FI.utf8
+   fr_FR.utf8
+   nb_NO.utf8
+   sv_SE.utf8
    ```
 
 ### Installation on FreeBSD
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -32,23 +32,7 @@ Docker are found there.
 The rest of this document is about doing a local installation of Zonemaster-CLI,
 not relevant for running Zonemaster-CLI on Docker.
 
-## Pre-built packages
-
-Pre-built packages are available for Debian and Ubuntu (architecture amd64 only).
-
-> Note: In order to use algorithm 15 and 16 on Ubuntu 20.04 the CPAN installation
-> must be used.
-
-1) Configure Zonemaster repository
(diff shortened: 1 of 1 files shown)


pnax:
Could you add an entry in the table of contents?

matsduf:
Milestone is not set on this PR. What is the intent?

docs/Installation.md
```diff
@@ -31,6 +32,22 @@ Docker are found there.
...
+
+1) Configure Zonemaster repository
+   ```sh
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo bash
+   ```
+2) Install Zonemaster CLI
+   ```sh
+   sudo apt install zonemaster-cli
+   ```
+
```
matsduf, line 50: 1. Will the package also update the locale setting?
2. Without support for algorithms 15 and 16 Zonemaster is incomplete. We should not have instructions for installing an incomplete Zonemaster. Furthermore, we state in the main README that there is support for algorithms 15 and 16 on Ubuntu 20.04.
3. I do not see why this is put in a separate section. Instead this should be the installation instruction for Debian.
4. The installation instruction for Engine for Debian should also be updated. Else you will first install Zonemaster::LDNS and Zonemaster::Engine with CPAN and come here to install the same thing with package.

docs/Installation.md
```diff
@@ -31,6 +32,22 @@ Docker are found there.
...
+## Pre-built packages
+
+Pre-built packages are available for Debian and Ubuntu (architecture amd64 only).
+
+> Note: In order to use algorithm 15 and 16 on Ubuntu 20.04 the CPAN installation
+> must be used.
+
+1) Configure Zonemaster repository
+   ```sh
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo bash
```
mattias-p, line 44: Maybe change `bash` to `sh`?

matsduf (changes requested):
If this instruction is based on packages, then the instruction for Engine for Debian should also be updated. Else you will go there first and install Zonemaster::LDNS and Zonemaster::Engine from CPAN and then come here and install those as packages also, maybe in different paths.

Is this for v2021.2? If so, does the release process need an update?

mattias-p (commented):
Maybe change the `Prerequisites` heading to `Prerequisites for CPAN installation`?

docs/Installation.md
blacksponge: > The installation instruction for Engine for Debian should also be updated. Else you will first install Zonemaster::LDNS and Zonemaster::Engine with CPAN and come here to install the same thing with package.

There is technically no need to go to install the engine / ldns first as it will be installed as dependency of the cli. I will update of users wanting the engine only after this PR as been done (the changes are very similar so once this one is resolved I can port them to engine)

> I do not see why this is put in a separate section. Instead this should be the installation instruction for Debian.

That is what I did first but then the instruction for packages are after the Prerequisites section even though there is no Prerequisites. So to make it clearer I preferred having two sections. Also this way it is clearer that a new installation method is available. I can put it back to the other section if you think that is better.

> Without support for algorithms 15 and 16 Zonemaster is incomplete. We should not have instructions for installing an incomplete Zonemaster. Furthermore, we state in the main README that there is support for algorithms 15 and 16 on Ubuntu 20.04.

I can make it Debian only if that is an issue, but we will be ofuscating the fact that packages will actually work on Ubuntu 21.04+ even though we don't officially support it.

> Will the package also update the locale setting?

No, I forgot about that.

docs/Installation.md
blacksponge: > Maybe change the Prerequisites heading to Prerequisites for CPAN installation?

Sure, I can change the Prerequisites header.

> Maybe change bash to sh?

I haven't tested with sh, I suspect it will work, if it does I will update.

docs/Installation.md
blacksponge: > Milestone is not set on this PR. What is the intent?

I set the milestone for the v2021.2. I will try to update the release process asap, if it is blocking can it be done in a fix release?

blacksponge:
I have updated the description with instruction to test that for the develop branch.

docs/Installation.md
matsduf: > There is technically no need to go to install the engine / ldns first as it will be installed as dependency of the cli. I will update for users wanting the engine only after this PR as been done (the changes are very similar so once this one is resolved I can port them to engine)

If you start from the main README then you come to Zonemaster-Engine first. That was my point.

matsduf: > I can make it Debian only if that is an issue, but we will be ofuscating the fact that packages will actually work on Ubuntu 21.04+ even though we don't officially support it.

It does not work with Ubuntu as we support Zonemaster. We do not want installations without support for algorithm 15 and 16.

matsduf: >> Will the package also update the locale setting?
>
> No, I forgot about that.

How do you plan to resolve that?

blacksponge: > How do you plan to resolve that?

I did not forgot to add it in the package, just in the documentation. In my opinion this should not be done by the package.

matsduf:
Installing with package is also local installation. Make installation with package on Debian the only option for Debian.

docs/Installation.md
blacksponge: > It does not work with Ubuntu as we support Zonemaster. We do not want installations without support for algorithm 15 and 16.

Then I will remove the mention for Ubuntu

matsduf: I am fine with having the locale installation outside the package.

blacksponge: > Installing with package is also local installation. Make installation with package on Debian the only option for Debian.

I'd prefer not, this won't work for other architecture that amd64 (if someone wants to install the cli on arm64 they would need the cpan instructions)

matsduf: Zonemaster is only supported on "x86_64 / amd64" 

https://github.com/zonemaster/zonemaster/tree/develop#supported-processor-architectures

matsduf: I could be fine to have both package and CPAN installation instructions for Debian, but integrate it into the "local installation".

blacksponge: > I could be fine to have both package and CPAN installation instructions for Debian, but integrate it into the "local installation".

Okay, I'll do that. Even if we don't support other architecture, I prefer to keep both at least for this release as this is quite new.

matsduf: Yes, that is a good point.

blacksponge: I drafted a release process here: https://github.com/zonemaster/zonemaster/pull/1019

blacksponge:
I think I have addressed most of the comments, please re-review

docs/Installation.md
```diff
@@ -31,7 +31,7 @@ Docker are found there.
 The rest of this document is about doing a local installation of Zonemaster-CLI,
 not relevant for running Zonemaster-CLI on Docker.
 
-## Prerequisites
+## Prerequisites for CPAN installation
 
 Before installing Zonemaster::CLI, you should [install Zonemaster::Engine][
 Zonemaster::Engine installation].
```
matsduf, line 37: When you come to the headline you do not even know that there is an alternative with package. I suggest the following:

```
## Prerequisites for CPAN installation

Before installing Zonemaster::CLI from CPAN, you should [install Zonemaster::Engine][
Zonemaster::Engine installation], unless you are to install on Debian using pre-built
packages (see below).
```

docs/Installation.md
```diff
@@ -91,6 +90,19 @@ Zonemaster::CLI, see the [declaration of prerequisites].
...
+2) Configure Zonemaster repository
+   ```sh
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo sh
+   ```
+3) Install Zonemaster CLI
+   ```sh
+   sudo apt install zonemaster-cli
+   ```
+
+#### CPAN Installation
```
matsduf, line 104: I think it would be better with
```
### Installation on Debian
Using pre-built packages is the preferred method for Debian. If you prefer to
install from CPAN instead, follow the steps for Ubuntu.

(Steps to install with packages)


### Installation on Ubuntu

(Steps to install from CPAN)
```

docs/Installation.md
blacksponge: done

docs/Installation.md
blacksponge: I have updated like you suggested. I also moved the locales configuration after installation because on minimal environment (I am testing in debian/ubuntu container) the `locales` packages in not installed, I added it to the dependencies.

docs/Installation.md
```diff
@@ -72,9 +74,20 @@ Zonemaster::CLI, see the [declaration of prerequisites].
...
+### Installation on Debian
 
-1) Update configuration of "locale"
+Using pre-built packages is the preferred method for Debian. If you prefer to
+install from CPAN instead, follow the steps for Ubuntu.
+
+1) Configure Zonemaster repository
+   ```sh
+   curl -Ls https://package.zonemaster.net/setup.sh | sudo sh
+   ```
```
matsduf, line 85: I think that "Configure Zonemaster repository" is confusing. What is updated is a the local list of repositories, not a Zonemaster repository.

I do not think that anyone should execute the command curl -Ls https://package.zonemaster.net/setup.sh | sudo sh. This something unknown that you will execute as root. I do not think we should have that in our instructions. It is better have an separate section at the end (appendix) with the commands. Instead we should comment that using package will make upgrade easier.

docs/Installation.md
blacksponge: I have updated with sh

docs/Installation.md
matsduf: I think it looks fine now except for that I am uncomfortable with the remote script. Is that only me?

docs/Installation.md
blacksponge: I can change it to:
Download install script
```
curl -Ls https://package.zonemaster.net/setup.sh
```
Review the content 
```
less setup.sh
```
Execute the install script
```
sudo sh setup.sh
````

Comment (blacksponge) [reply docs/Installation.md:85]: > Configure Zonemaster repository
I can change it to `Configure Zonemaster packages repository` if that is better

Comment (blacksponge) [reply docs/Installation.md:85]: >I do not think that anyone should execute the command

Otherwise the instruction would be (master branch only)

```
curl -Ls https://package.zonemaster.net/zonemaster-archive-keyring.gpg | sudo tee /usr/share/keyrings/zonemaster-archive-keyring.gpg > /dev/null

. /etc/os-release

cat <<EOF | sudo tee /etc/apt/sources.list.d/zonemaster.sources
Types: deb deb-src
URIs: https://package.zonemaster.net/debian/
Suites: ${VERSION_CODENAME}
Architectures: $(dpkg --print-architecture)
Components: main
Signed-By: /usr/share/keyrings/zonemaster-archive-keyring.gpg
EOF

sudo apt update
```

File: docs/Installation.md (hunk 4 above)
Comment (blacksponge) [reply docs/Installation.md]: I have shared opinion on the subject, but I think it has become quite a popular pattern (even if insecure) for third party package installation. As stated in the other comment if that is better I can split the command in two parts (download and then execute, as opposed to piping to sh), which is a bit better in my opinion.

Comment (blacksponge) [reply docs/Installation.md]: In the end I think short and simple instructions are nice to have and can bring new users that don't want to deal with longer  or more complicated ones.

**Review by pnax — APPROVED** [review]
Nice! I've tested it and this looks fine for me. I think we can start with this proposal and improve it during next development cycle if needed.

@matsduf Would it be possible to add an entry in the CLI README document stating that we support packaging for Debian (as done with Docker)?

And about the remote script piped to sh, I'd rather avoid it as well. However I'm fine having it, since I really like the support for Debian package.

File: docs/Installation.md (hunk 5 above)
Comment (matsduf) [reply docs/Installation.md]: I guess an alternative is `sudo dpkg -i package_file.deb` but the difference is minor, I guess, because then you trust that the deb package does not execute evil code.

File: docs/Installation.md [hunk 6]
```diff
@@ -72,9 +74,20 @@ Zonemaster::CLI, see the [declaration of prerequisites].
...
 
 
-### Installation on Debian or Ubuntu
+### Installation on Debian
 
-1) Update configuration of "locale"
+Using pre-built packages is the preferred method for Debian. If you prefer to
+install from CPAN instead, follow the steps for Ubuntu.
+
+1) Configure Zonemaster packages repository
```
matsduf, line 82: Is it correct to call this "Configure Zonemaster packages repository"? 
* "Add Zonemaster packages repository to repository list"
* "Configure OS packages repository list"

docs/Installation.md
blacksponge: I honestly don't see much difference between downloading and installing a deb file or downloading and executing a sh script. I thought about that alternative too, but I don't think it is worth the extra work.

docs/Installation.md
blacksponge: "Add Zonemaster packages repository to repository list" is nice

matsduf:
> @matsduf Would it be possible to add an entry in the CLI README document stating that we support packaging for Debian (as done with Docker)?

It could be added, but I am not sure it is needed. You want to do a local installation on Debian, and then you are instructed to use packages. It will be in the release notes, though.

matsduf (dismissed):
I have one remaining comment. If you update based on that I will re-approve.

I approve.

docs/Installation.md
blacksponge: changed

blacksponge:
> I have one remaining comment. If you update based on that I will re-approve.

I also splited the `curl | sh` in two parts, giving the possibility for the users to review the script (and making sure that what they download is what they execute).

matsduf (approved):
I approve.

matsduf:
> I also splited the `curl | sh` in two parts, giving the possibility for the users to review the script (and making sure that what they download is what they execute).

I like that.

blacksponge:
I went a bit too fast and forgot a curl option

matsduf (approved):
I approve.

matsduf:
For this to work well, the instructions for Zonemaster-Engine should also be updated.

pnax (approved):

docs/Installation.md
mattias-p: Maybe we could let the user verify a checksum of the script? Would that improve the security situation enough to be worth the extra work for our users and ourselves?

mattias-p (approved):

docs/Installation.md
matsduf: It maybe improves the security slightly, but I do not think it is worth it.

PR #234, opened 2021-12-03
Preparation for release v2021.2 (CLI)

matsduf:
## Purpose

* Bumps version in perl module (compared to last release)
* Updates Changes file

At last relase VERSION was `v3.1.0`. During development this has been increased to `v3.5.0` but for this release it should really be `v3.1.1` which it has been reset to in this PR.

Set this version to require latest version of Engine, which is the version that tests have run against.

## Changes

Updated files:

* Changes
* lib/Zonemaster/CLI.pm
* Makefile.PL

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,23 @@
 Release history for Zonemaster component Zonemaster-CLI 
 
+v3.1.1 2021-12-03 (public release version)
+
+ [Features]
+ - Deb packages are available for Debian (#233)
+ - Adds support for Docker (#221, #218, #216)
+ - Replaces CentOS with Rocky Linux (#220)
+
+ [Fixes]
+ - Updates Finnish translation (#230, #224)
+ - Updates Norwegian translation (#228, #226)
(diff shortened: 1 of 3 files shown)


mattias-p (approved):
There are slight grammatical differences among the items, but I don't think that's a huge problem.

mattias-p:
Come to think of it, do we have reason to believe that this version of CLI wouldn't work with Engine 4.002? Otherwise I don't think it's a great idea to bump the Engine version requirement.

matsduf:
> Come to think of it, do we have reason to believe that this version of CLI wouldn't work with Engine 4.002? Otherwise I don't think it's a great idea to bump the Engine version requirement.

That is to make an update of CLI also update Engine to include the fixes in Engine. Also that is the version that we have tested CLI against. Furthermore, I have just followed the instructions in the release process.

mattias-p:
I created zonemaster/zonemaster#1022 to update the release process instruction.

PR #238, opened 2021-12-15
Clarifies IPv6 limitations

matsduf:
## Purpose

If zonemaster-cli is run in a non-IPv6 setting then meaningless error messages are outputted unless IPv6 has been turned off for zonemaster-cli. Docker is usually non-IPv6.

This PR updates the document to clarify this.

## How to test this PR

This documentation only. Follow the instructions to see the behavior.

### DIFF
--- a/USING.md
+++ b/USING.md
@@ -27,10 +27,17 @@ on the computer and that you can run Docker on it.
 When Docker has been correctly installed, no more installation is needed to run
 `zonemaster-cli`. Just follow the examples below.
 
+There is a limitation in Docker regarding IPv6. Unless IPv6 has been enabled in
+the Docker daemon, there is no support for IPv6. To avoid meaningless errors,
+use `--no-ipv6` if there is no IPv6 support. Also see "[Enable IPv6 support]".
+
 ### Local installation
 
 To have an local installation of Zonemaster-CLI follow the
 [installation instruction]. When installed, the examples below can be followed.
+
+If the network has no IPv6 support (common in home networks) then turn off IPv6.
+Use `--no-ipv6` to avoid meaningless errors if there is no IPv6 support.
 
 
 ## Invoking the command line tool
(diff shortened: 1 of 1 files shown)


matsduf:
I think it would be good to include this in the v2021.2.1 patch release. If you disagree I will move it to v2022.1.

mattias-p (approved):
The additions to the installation sections are great. But I feel the additions to the Invocation section are a bit excessive. The `--no-ipv6` option is already explained both above and below. From my point of view it would be better to let the Invocation section focus on getting things done.

pnax (approved):
> From my point of view it would be better to let the Invocation section focus on getting things done.

I have the same point of view :)

matsduf:
@mattias-p and @pnax, after seeing the reaction of some pilot users of zonemaster-cli on Docker I think that it is needed to lift `--no-ipv6` up, and some people just jump to the invocation section and skip other parts. But I do agree that the text was too wordy, so I have reduced that somewhat. I hope that you find that to be an improvement. Please re-approve.

USING.md
```diff
@@ -27,18 +27,29 @@ on the computer and that you can run Docker on it.
...
 
-With Docker:
+Using Docker:
+```sh
+docker run -t --rm zonemaster/cli zonemaster.net --no-ipv6
+```
+or
 ```sh
 docker run -t --rm zonemaster/cli zonemaster.net
 ```
```
pnax, line 55: Since I understand Docker is not configured with IPv6 by default, I think we can leave only one command that most users will use (the one with `--no-ipv6`) and remove the second one (letting people remove the option if needed).

USING.md
```diff
@@ -47,6 +58,10 @@ With local installation:
 ```sh
 zonemaster-cli zonemaster.net
 ```
+or
+```sh
+zonemaster-cli zonemaster.net --no-ipv6
+```
```
pnax, line 64: Again, I would suggest to only keep the default (and possibly most used) command and let users tweak it to their needs.

pnax (commented):
Nice, I would even suggest to go one step further and only keep the command that should work out of the box.

mattias-p (approved):

USING.md
matsduf: Can we leave as-is for now, and tweak for next release?

USING.md
matsduf: Can we leave as-is for now, and tweak for next release? If so, can you approve?

pnax (approved):

PR #241, opened 2021-12-20
Includes information on --pull for Docker

matsduf:
## Purpose

When an image has been downloaded Docker will normally never look for updated image if the same tag is used. There is a risk that a user will never get an update of Zonemaster. The document has been updated to suggest that `--pull` is used at least sometimes to pull updates.

## Changes

Updated files:
* USING.md

## How to test this PR

Follow the instructions in the document to make sure they are correct.

### DIFF
--- a/USING.md
+++ b/USING.md
@@ -2,7 +2,9 @@
 
 ## Table of contents
 * [Docker or local installation](#Docker-or-local-installation)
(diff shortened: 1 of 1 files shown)


mattias-p (approved):
LGTM

PR #253, opened 2022-04-28
remove domains final dots

blacksponge:
## Purpose

Do not crash when a domain is specified with a final dot.

## Context

https://github.com/zonemaster/zonemaster-engine/issues/1055

## Changes

* Remove final dot from domain if not root
* Remove final dot from NS

## How to test this PR

* `../zonemaster-cli/script/zonemaster-cli --ns ns2.nic.fr. zonemaster.net. --show-testcase` should run the test without crashing
* `../zonemaster-cli/script/zonemaster-cli --ns ns2.nic.fr.. zonemaster.net. --show-testcase` should show an error
* `../zonemaster-cli/script/zonemaster-cli --ns ns2.nic.fr. zonemaster.net.. --show-testcase` should show an error

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -472,6 +472,8 @@ sub run {
         die __( "Must give the name of a domain to test.\n" );
     }
 
+    $domain =~ s/\.$// unless $domain eq '.';
+
     if ( $translator ) {
         if ( $self->time ) {
             print __( 'Seconds ' );
@@ -601,6 +603,8 @@ sub add_fake_delegation {
             exit( 1 );
         }
 
+        $name =~ s/\.$//;
+
         if ($ip) {
             push @{ $data{ $self->to_idn( $name ) } }, $ip;
         }
@@ -704,7 +708,7 @@ sub print_test_list {
 
 sub do_dump_profile {
     my $json = JSON::XS->new->canonical->pretty;
-    
+
     print $json->encode( Zonemaster::Engine::Profile->effective->{ q{profile} } );
 
     exit;
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -603,7 +603,7 @@ sub add_fake_delegation {
             exit( 1 );
         }
 
-        $name =~ s/\.$//;
+        $name =~ s/\.$// unless $name eq '.';;
 
         if ($ip) {
             push @{ $data{ $self->to_idn( $name ) } }, $ip;
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -608,7 +608,7 @@ sub add_fake_delegation {
         }
 
         if ( $name =~ m/\.\./i ) {
-            say STDERR __x( "The name of the nameserver '{ns}' contains consecutive dots.", ns => $name );
+            say STDERR __x( "The name of the nameserver '{nsname}' contains consecutive dots.", nsname => $name );
             exit ( 1 );
         }
 
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -472,6 +472,10 @@ sub run {
         die __( "Must give the name of a domain to test.\n" );
     }
 
+    if ( $domain =~ m/\.\./i ) {
+        die __( "The domain name contains consecutive dots.\n" );
+    }
+
     $domain =~ s/\.$// unless $domain eq '.';
 
     if ( $translator ) {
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/CLI.pm
```diff
@@ -601,6 +603,8 @@ sub add_fake_delegation {
             exit( 1 );
         }
 
+        $name =~ s/\.$//;
```
tgreenx, line 606: Can the root be a NS in the context of fake delegation? I think it can, according to code I see in Engine ([here](https://github.com/zonemaster/zonemaster-engine/blob/master/lib/Zonemaster/Engine.pm#L112-L114)). @matsduf can you confirm?
Thus it would require `unless $name eq '.';` here as well.

lib/Zonemaster/CLI.pm
```diff
@@ -601,6 +603,8 @@ sub add_fake_delegation {
             exit( 1 );
         }
 
+        $name =~ s/\.$// unless $name eq '.';;
```
pnax, line 606: typo: two semi-colons

pnax (approved):

lib/Zonemaster/CLI.pm
matsduf: Yes, an NS can point at a name server name ".".

lib/Zonemaster/CLI.pm
matsduf: To resolve zonemaster/zonemaster-engine#1055 the code must remove multiple trailing dots. Cf https://github.com/zonemaster/zonemaster-backend/pull/983#discussion_r862703867

matsduf (changes requested):

lib/Zonemaster/CLI.pm
```diff
@@ -472,6 +472,8 @@ sub run {
         die __( "Must give the name of a domain to test.\n" );
     }
```
matsduf, line 474: I suggest adding:
```
die __( "Domain name must not end with double dots.\n" ) if $domain =~ s/[.]{2,}$//;
```

lib/Zonemaster/CLI.pm
```diff
@@ -601,6 +603,8 @@ sub add_fake_delegation {
             exit( 1 );
         }
```
matsduf, line 605: I suggest adding:
```
die __( "Name server name must not end with double dots.\n" ) if $name =~ s/[.]{2,}$//;
```

matsduf (changes requested):

lib/Zonemaster/CLI.pm
blacksponge: > typo: two semi-colons

fixed

blacksponge:
I took the same approach as https://github.com/zonemaster/zonemaster-backend/pull/983 and return an error if consecutive dots are found in names.

lib/Zonemaster/CLI.pm
```diff
@@ -597,10 +603,17 @@ sub add_fake_delegation {
         my ( $name, $ip ) = split( '/', $pair, 2 );
 
         if ( not $name ) {
-            say STDERR "--ns must have be a name or a name/ip pair.";
+            say STDERR __( "--ns must have be a name or a name/ip pair." );
```
pnax, line 606: weird wording, maybe remove "have"

blacksponge: Indeed, should be fixed

pnax (dismissed):

lib/Zonemaster/CLI.pm
```diff
@@ -597,10 +603,17 @@ sub add_fake_delegation {
         my ( $name, $ip ) = split( '/', $pair, 2 );
 
         if ( not $name ) {
-            say STDERR "--ns must have be a name or a name/ip pair.";
+            say STDERR __( "--ns must be a name or a name/ip pair." );
             exit( 1 );
         }
 
+        if ( $name =~ m/\.\./i ) {
+            say STDERR __x( "The name of the nameserver '{ns}' contains consecutive dots.", ns => $name );
```
matsduf, line 611: In a msgid in Zonemaster-Engine, an argument for this type must be "nsname". It would be good to use the same definitions here because it makes it easier for the translator. See https://github.com/zonemaster/zonemaster-engine/blob/develop/docs/logentry_args.md

matsduf (changes requested):
I suggest, in general, to handle the update of the PO files in separate PRs. New PO files can easily be generated with the tools in the share directory.

blacksponge:
@matsduf please re-review

blacksponge:
@matsduf please re-review

matsduf (approved):

pnax (approved):

PR #254, opened 2022-04-28
Use libidn2 instead of libidn

pnax:
## Purpose

Replace libidn with libidn2.

## Context

https://github.com/zonemaster/zonemaster-ldns/pull/133

## Changes

* Update Travis dependency
* Replace references to "libidn" with "libidn2"

## How to test this PR

Using Zonemaster::LDNS (with libidn2 support), Zonemaster should work the same as before.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -669,7 +669,7 @@ sub to_idn {
         return Zonemaster::LDNS::to_idn( decode( $self->encoding, $str ) );
     }
     else {
-        say STDERR __( "Warning: Zonemaster::LDNS not compiled with libidn, cannot handle non-ASCII names correctly." );
+        say STDERR __( "Warning: Zonemaster::LDNS not compiled with libidn2, cannot handle non-ASCII names correctly." );
         return $str;
     }
 }
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ addons:
     - libdevel-checklib-perl
     - libfile-sharedir-perl
     - libfile-slurp-perl
-    - libidn11-dev
+    - libidn2-dev
     - libintl-perl
     - libjson-pp-perl
     - liblist-moreutils-perl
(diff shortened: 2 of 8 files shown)


Changes
```diff
@@ -226,7 +226,7 @@ v1.1.0 2017-04-04 Public pre-release
 
 1.0.1    2015-04-07
  - State clearly that a given name is not a domain when it is not a domain
- - Send "Net::LDNS not compiled with libidn" to STDERR, not STDOUT
```
blacksponge, line 229: That should not be changed.

blacksponge (commented):
Just one comment, otherwise lgtm

Changes
pnax: Oups ^^'

blacksponge (approved):

Changes
pnax: fixed (hopefully I haven't done the same mistake in the other repos)

lib/Zonemaster/CLI.pm
```diff
@@ -669,7 +669,7 @@ sub to_idn {
         return Zonemaster::LDNS::to_idn( decode( $self->encoding, $str ) );
     }
     else {
-        say STDERR __( "Warning: Zonemaster::LDNS not compiled with libidn, cannot handle non-ASCII names correctly." );
+        say STDERR __( "Warning: Zonemaster::LDNS not compiled with libidn2, cannot handle non-ASCII names correctly." );
```
matsduf, line 672: I think the following would be better and not bound to details not relevant here.
```
Warning: Zonemaster::LDNS not compiled with IDN support, cannot handle names with non-ASCII characters correctly.
```
But I accept as-is.

matsduf (approved):

lib/Zonemaster/CLI.pm
pnax: Good point. I'll do the change in another PR.

Issue #252, opened 2022-03-30
Add global cli.args file (configuration file)


matsduf:
Zonemaster-cli has the feature that allows the user to create a configuration file (~/.zonemaster/cli.args) with default command line options:
```
CONFIGURATION
       If there is a readable file .zonemaster/cli.args in the user's home directory, each line in that file will be
       prepended as an argument on the command line. For example, if one would like to by default run with the log level
       set to DEBUG and with translation to human-readable messages turned off, one could put this in the config file:

          --raw
          --level=DEBUG

       Any arguments actually given on the command line will override what's in the config file.
```
(from `man zonemaster-cli`)

Add the support for a global configuration file (/etc/zonemaster/cli.args ; /usr/local/etc/zonemaster/cli.args for FreeBSD) that works in the same way.
* Any options in the private configuration file should overwrite options in the global configuration file.
* Any options on the command line should also overwrite.

The new feature was requested in an email to the zonemaster-users list by @aabdnn.

matsduf:
@aabdnn, could you provide a description of your need for this feature as a comment to this issue?

mattias-p:
Oh. I wasn't aware of the discussion on the mailing list.

anandb-ripencc:
My use case is that we have a Zonemaster server, used by multiple users to do checks. Some use the web gui, and some use the cli, but cli users are at a disadvantage, because they would all have to remember that we have a custom profile and use it.

I am aware of `~/.zonemaster/cli.args`, which can be used to specify custom arguments for the cli. However, this is per-user. If this scheme were extended to instead (or additionally) read `/etc/zonemaster/cli.ini` (similar to how the backend reads backend_config.ini), then a custom profile can be defined there globally for all cli users.

mattias-p:
Thank you so much for explaining! I had no idea Zonemaster was being used like this.

matsduf:
Resolved by #260.

PR #260, opened 2022-05-11
Adds global configuration file

matsduf:
## Purpose

zonemaster-cli has support for a configuration file, cli.args, for each user, with the same arguments as on the command line. #252 requests a global configuration file. This PR adds a possibility to have such a file.

The global configuration file is hard coded to be /etc/zonemaster/cli.args (Linux style) or /usr/local/etc/zonemaster/cli.args (FreeBSD style), i.e. the same directory as the Backend configuration is stored in by default.

Arguments on the command line takes precedence over the equivalent argument in ~/.zonemaster/cli.args, which takes precedence over the equivalent argument in a global cli.args. Only one global file is used, /etc/zonemaster/cli.args if that exists, else /usr/local/etc/zonemaster/cli.args.

The way the cli.args is read has been updated to support empty lines and lines starting with "#", comments.

## How to test this PR

Create ~/.zonemaster/cli.args, /etc/zonemaster/cli.args and /usr/local/etc/zonemaster/cli.args and verify that arguments are used and the the precedence order works.

### DIFF
--- a/script/zonemaster-cli
+++ b/script/zonemaster-cli
@@ -7,50 +7,50 @@ use Zonemaster::CLI;
 use File::Spec;
 use autodie;
 
-# Load default arguments from file in home directory, if any
-# This must be loaded before any global file to make the local
-# file take precedence
-my $homedir  = ((getpwuid($<))[7]) || $ENV{HOME};
-my $homeconffile = File::Spec->catfile($homedir, '.zonemaster', 'cli.args');
-
-if (-r $homeconffile) {
-    open my $fh, '<', $homeconffile;
+sub read_conf_file {
+    # Returns list of command line parameters. List can be empty.
+    my ($conf_file) = @_;
     my @lines;
+    open my $fh, '<', $conf_file;
     while (<$fh>) {
         chomp;
         next if /^\s*$/;
         next if /^\s*#/;
         push @lines, $_;
     };
+    return @lines;
+}
+
+# Load default arguments from file in home directory, if any
+# This must be loaded before any global file to make the local
+# file take precedence
+my $home_dir  = ((getpwuid($<))[7]) || $ENV{HOME};
+my $home_conf_file = File::Spec->catfile($home_dir, '.zonemaster', 'cli.args');
+
+if (-r $home_conf_file) {
+    my @lines = read_conf_file ($home_conf_file);
     unshift @ARGV, @lines;
 }
 
 # Load default arguments from global file, if any
-my @globalconf = (
+my @global_conf = (
     '/etc/zonemaster/cli.args',
     '/usr/local/etc/zonemaster/cli.args'
     ); # Order is significant.
-my $globalconffile;
+my $global_conf_file;
 
-for my $p (@globalconf) {
+for my $p (@global_conf) {
     if ( -e $p and -r $p ) {
-        $globalconffile = $p;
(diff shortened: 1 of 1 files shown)


matsduf:
@aabdnn, this is a proposal to give global configuration file (cli.args) for zonemaster-cli.

script/zonemaster-cli
```diff
@@ -7,16 +7,51 @@ use Zonemaster::CLI;
...
-    chomp(@lines);
+if ( defined $globalconffile ) {
+    open my $fh, '<', $globalconffile;
+    my @lines;
+    while (<$fh>) {
+        chomp;
+        next if /^\s*$/;
+        next if /^\s*#/;
+        push @lines, $_;
+    };
```
pnax, line 50: Did you consider making a function to avoid code duplication when reading each file?

matsduf: Fixed.

matsduf:
@mattias-p and @pnax, please re-review.

mattias-p:
I came to think about something. Global configuration of command line parameters could be achieved with zero lines of code in zonemaster-cli by adding an alias to /etc/profile. I suggest this PR is updated to simply explain how this is done in the "Configuration" POD section.

A possible further step in this direction would be to deprecate the cli.args feature and instruct users to replace their cli.args file with an alias in $HOME/.profile.

When we finally drop the cli.args feature:
* We'll have a uniform format for both system-wide and per-user configuration of command line parameters.
* We'll impose one one fewer configuration formats on the world. (The use of alias is already established.)
* We achieve this in a negative number of lines of code. Meaning less room for bugs.

Thoughts on this?

matsduf:
> I came to think about something. Global configuration of command line parameters could be achieved with zero lines of code in zonemaster-cli by adding an alias to /etc/profile. I suggest this PR is updated to simply explain how this is done in the "Configuration" POD section.

A global alias in /etc/profile works for those that use SHELL that uses `/etc/profile`. It could require that aliases are created in different files for different SHELL.

> A possible further step in this direction would be to deprecate the cli.args feature and instruct users to replace their cli.args file with an alias in $HOME/.profile.

Not all SHELL uses .profile so the instruction would be more complex.

> Thoughts on this?

I find the solution with `cli.args` to be more robust and user friendly. The code needed is not complex and it is easy to test. There is a clear use cases for both private and global `cli.args`.

mattias-p:
> A global alias in /etc/profile works for those that use SHELL that uses /etc/profile. It could require that aliases are created in different files for different SHELL.

> Not all SHELL uses .profile so the instruction would be more complex.

All shells in the Bourne shell family use /etc/profile and .profile.

Notably the Csh family does not use these particular files. Though surely anyone who uses one of those shells is an advanced user how has no trouble configuring their system-wide and per-user aliases even when the documentation only refers /etc/profile and .profile.

matsduf:
Csh is the default shell for root user in FreeBSD.

It is also simpler to handle a cli.arg file since changes take effect directly without having any special reload. Also update of cli.args does not effect anything else.

matsduf:
@pnax and  @mattias-p, please re-review.

pnax (approved):
lgtm

mattias-p (approved):

matsduf:
@aabdnn, it will be included in the upcoming release.

Issue #277, opened 2022-07-07
How to find line feed issues in PO file?


matsduf:
Note, this issue is about the general case of discovering the errors, not about the specific error.

The issue described here is probably applicable for Zonemaster-Engine and Zonemaster-Backend too.

When building a new distribution and after running "make all" I got:
```
cp lib/Zonemaster/CLI.pm blib/lib/Zonemaster/CLI.pm
cp script/zonemaster-cli blib/script/zonemaster-cli
"/usr/local/bin/perl" "-Iinc" -MExtUtils::MY -e 'MY->fixin(shift)' -- blib/script/zonemaster-cli
cd share && "gmake" all
gmake[1]: Entering directory '/usr/home/matsd/git/zonemaster-cli/share'
da.po:180: 'msgid' and 'msgstr' entries do not both end with '\n'
msgfmt: found 1 fatal error
gmake[1]: *** [GNUmakefile:27: da.mo] Error 1
gmake[1]: Leaving directory '/usr/home/matsd/git/zonemaster-cli/share'
*** Error code 2

Stop.
make: stopped in /usr/home/matsd/git/zonemaster-cli
```

The following was the error:
```
msgid "The domain name contains consecutive dots.\n"
msgstr "Domænenavnet indeholder flere dots (.) efter hinanden."
```

I found the following where msgid ends with two line feed, but mgstr only one. But that does not create an error.

```
msgid ""
"Warning: setting locale category LC_MESSAGES to %s failed (is it installed "
"on this system?).\n"
"\n"
msgstr ""
"Advarsel: Kunne ikke sætte locale kategori LC_MESSAGES til %s (er det "
"installeret på dette system?).\n"
```

There should be a tool to make it possible to discover this early. Is there?

pnax:
Done with #314.

PR #278, opened 2022-07-07
Fixes error in PO file that prevents MO to be created

matsduf:
## Purpose

PO file cannot be compiled into MO file due to a bug.

## Context

#277 describes the error, but also the problem how to discovering it early.

## Changes

Fixes the da.po.

## How to test this PR

Run the distribution process and make sure "make all" works in the root of the respository.


mattias-p (approved):
Looking good.

PR #285, opened 2022-10-22
List MooseX::Getopt::GLD as prereq, not MooseX::Getopt

haarg:
This dist uses MooseX::Getopt::GLD, not MooseX::Getopt.

MooseX::Getopt::GLD is currently packaged with MooseX::Getopt, but should still be listed as a prereq on its own.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -20,7 +20,7 @@ tests_recursive( 't' );
 requires(
     'JSON::XS'           => 0,
     'Locale::TextDomain' => 1.23,
-    'MooseX::Getopt'     => 0,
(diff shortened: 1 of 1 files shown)


matsduf:
Thanks, we will take a look at it.

haarg:
Generally, you can't trust that a module will continue to be part of the same distribution. Distributions can be split, and if you aren't depending on the actual modules you are using, your dependencies will be broken.

For this dist it's probably not really a problem, since MooseX::Getopt is pretty much documented as being an alias for MooseX::Getopt::GLD. I just noticed this when chasing some unrelated issues. I'm not really bothered if this isn't merged.

PR #287, opened 2022-11-04
Fail multiple domain names in entry

### LINKED ISSUE #286: Make zonemaster-cli complain over multiple domain names in entry
Domain (zone) name is mandatory for `zonemaster-cli` it is given as a single string without any `--<OPTION>` before. If multiple domain names are given the first one is tested and the others are just ignored with no error or message:
```
# zonemaster-cli iis.se telia.com
Seconds Level     Message
======= ========= =======
   6.53 NOTICE    SOA 'mname' nameserver (ns.nic.se) is not listed in "parent" NS records for tested zone (nsa.dnsnode.net;nsp.dnsnode.net;nsu.dnsnode.net).
```
In the example above it is maybe easy to see, but the command line could be very long, e.g.
```
zonemaster-cli iis.se --test connectivity/connectivity02 --level debug --show-testcase --raw --ns nsa.dnsnode.net/194.58.192.46 ns1.telia.com/10.10.10.10
```
and it could be hard to see that `--ns` is missing for the second name server.

Make CLI.pm check for multiple strings not specified with `--<OPTION>`

matsduf:
## Purpose

It is possible to add several domain names on the command line for `zonemaster-cli` but only the first one is used. The others are ignored. This PR changes the behavior and makes it fail if more than one domain is provided.

## Context

Fixes #286.

This change could be considered to be a bug fix or a change of the API. Maybe the behavior was on purpose.

## How to test this PR

Run `zonemaster-cli` with more than one domain name on the command line. It should fail:
```
$ zonemaster-cli iis.se telia.com
Only one domain can be given for testing. Did you forget to prepend an option with '--<OPTION>'?
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -474,6 +474,10 @@ sub run {
         print "\n" if $fh_diag eq *STDOUT;
     }
 
+    if ( scalar @{ $self->extra_argv } > 1 ) {
+        die __( "Only one domain can be given for testing. Did you forget to prepend an option with '--<OPTION>'?\n" );
+    }
+
     my ( $domain ) = @{ $self->extra_argv };
     if ( not $domain ) {
         die __( "Must give the name of a domain to test.\n" );


marc-vanderwal (approved):
Looks good to me.

And I’d see it as both a bug fix and an API change.

tgreenx (approved):

PR #298, opened 2023-01-23
Update French translation

pnax:
## Purpose

This PR updates French translations.

## Context

n/a

## Changes

fr.po

## How to test this PR

Review translated strings.


share/fr.po
```diff
@@ -72,6 +72,10 @@ msgstr ""
 "Un nom ou une adresse IP correspondant à un serveur de noms pour les tests "
 "sur des zones non déléguées. Peut être utilisée plusieurs fois."
 
+msgid "Name of a root hints file to override the defaults."
+msgstr "Chemin du fichier d'indices de la racine remplaçant le fichier par "
```
marc-vanderwal, line 76: I think “root hints” is better translated as “indications de la racine”. See for example https://www.bortzmeyer.org/8499.html.

share/fr.po
```diff
@@ -200,6 +204,13 @@ msgstr ""
 "--level doit avoir pour valeur CRITICAL, ERROR, WARNING, NOTICE, INFO, "
 "DEBUG, DEBUG2 ou DEBUG3.\n"
 
+msgid ""
+"Only one domain can be given for testing. Did you forget to prepend an "
+"option with '--<OPTION>'?\n"
+msgstr ""
+"Uniquement un domaine peut être testé. Auriez-vous oublié de préfixer une "
```
marc-vanderwal, line 211: How about “Seul un domaine peut être soumis à un test”?

share/fr.po
pnax: Thanks for the suggestion. Fixed.

share/fr.po
pnax: Updated.

marc-vanderwal (approved):
Looks good to me.

Issue #181, opened 2020-11-18
Version of Zonemaster::LDNS


pawal:
The CLI --version does not show which version of Zonemaster::LDNS is being used. Which is more important than Module versions, since they are all included in the specific Engine version.

```
$ zonemaster-cli --version
CLI version:    v3.0.1
Engine version: v4.0.3

Test module versions:
        Address: v1.0.8
        Basic: v1.0.15
        Connectivity: v1.0.17
        Consistency: v1.1.15
        DNSSEC: v1.1.25
        Delegation: v1.0.19
        Nameserver: v1.0.26
        Syntax: v1.0.9
        Zone: v1.0.14
```

matsduf:
It is a good suggestion to add the version of Zonemaster::LDNS, and that should not be too complicated. @pawal, do you suggest to remove the module versions, or do you think they should be kept?

pawal:
I personally don't see a use from them, don't know what others think. There is no clear use-case for it to be included as I see it.

pnax:
Done in #306

PR #306, opened 2023-02-14
Zonemaster::LDNS and libldns versions

pnax:
## Purpose

Show which versions of Zonemaster-LDNS and NL NetLabs LDNS are being used.
Remove module versions.

## Context

Addresses https://github.com/zonemaster/zonemaster-cli/issues/181

## Changes

CLI.pm and version() method.

## How to test this PR

Verify that the following command returns 2 new lines `LDNS version` and `libldns version`:
```
$ zonemaster-cli --version
Zonemaster-CLI version v5.0.1                                                                                          
Zonemaster-Engine version v4.6.1     
Zonemaster-LDNS version 3.1.0                    
NL NetLabs LDNS version 1.8.3
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -673,13 +673,6 @@ sub print_versions {
     say 'Engine version: ' . $Zonemaster::Engine::VERSION;
     say 'LDNS version: ' . $Zonemaster::LDNS::VERSION;
     say 'libldns version: ' . Zonemaster::LDNS::lib_version();
-    say "\nTest module versions:";
-
-    my %methods = Zonemaster::Engine->all_methods;
-    foreach my $module ( sort keys %methods ) {
-        my $mod = "Zonemaster::Engine::Test::$module";
-        say "\t$module: " . $mod->version;
-    }
 
     return;
 }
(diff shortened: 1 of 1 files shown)


matsduf:
> Show which versions of Zonemaster::LDNS and libldns are being used.

That is a good idea.

> ```
> $ zonemaster-cli --version
> CLI version:    v5.0.1
> Engine version: v4.6.1
> LDNS version: 3.1.0
> libldns version: 1.8.3
> ...
> ```

However, I think that more updates should be done. Just as the issue points out the module numbers do not add anything. I suggest that they should be removed.

Secondly, we should use the full name of the Zonemaster components, especially it is problematic to refer to "LDNS" when we mean Zonemaster-LDNS since LDNS is a different code. Compare with how it is displayed when running the web GUI.

Thirdly, I think it would be better to refer to the NL NetLabs code as LDNS since that is what they call it. I suggest that "zonemaster-cli --version" outputs the following (and nothing else):

```
$ zonemaster-cli --version
Zonemaster-CLI version v5.0.1
Zonemaster-Engine version v4.6.1
Zonemaster-LDNS version 3.1.0
NL NetLabs LDNS version 1.8.3
```

pnax:
Updated to address your remarks.

matsduf (approved):

matsduf:
I also fee this is "Minor", but technically, isn't this a change of API, i.e. "Major"?

Issue #247, opened 2022-01-07
Clarify that --raw, --json, --json-stream and --json-translate


matsduf:
The options --raw, --json and --json-stream give alternative output of the test case results. If none of them are given, the default output is provided. If more than one of them are given, one is respected and the other or others are ignored.

The option --json-translate can only be used in combination with --json-stream, else it is ignored.

1. Clarify the documentation that three formats cannot be combined.
2. Make the code raise an exception if format options are combinded.
3. Make the code raise an exception if --json-translate is used without --json-stream

Input from #246.

mattias-p:
I suggest that we introduce a `--mode=MODE` option that takes the values `raw`, `json` and `json-stream`. The original options could be deprecated and kept around for a while, but defined to just set the value of the `--mode` option and they would not be allowed to be specified together with the `--mode` option.

matsduf:
There are pros and cons. With "mode" it is probably easier to see how the different output formats relate to each other, but "--raw" is shorter and simpler to write than "--mode=raw", and probably easier to remember.

Secondly, there is also the "--json-translate" to be handled in a good way.

Is there a good use case for both "json" and "json-stream"? Is there a good reason why "translate" is only implemented with "json-stream" and not "json"?

mattias-p:
Translation should probably be implemented for both `json` and `json-stream`.

I guess it's useful to have both `json` and `json-stream` for integration. `json-stream` is better for stream processing but some tools might not support it. I don't have any examples though.

Can you expand on what you mean with --json-translate being handled in a good way?

Oh, and there's also the `dump-profile` mode we need to consider. Another possibility is to use sub-commands instead of a --mode option to control the execution mode. `zonemaster-cli raw` is even shorter than `zonemaster-cli --raw`.

matsduf:
> Translation should probably be implemented for both `json` and `json-stream`.

It is called "translation" but what happens is that an extra field "message" is added. It could be added by default in both cases, and then deprecate "--json-translate".

> Can you expand on what you mean with --json-translate being handled in a good way?

Nothing more than the description says. "--json-translate" is silently ignored if "--json-stream" is not selected.

> Oh, and there's also the `dump-profile` mode we need to consider. Another possibility is to use sub-commands instead of a --mode option to control the execution mode. `zonemaster-cli raw` is even shorter than `zonemaster-cli --raw`.

Today a bareword is taken to be a domain name. If the domain name is "raw" how should the command be?

[...]

PR #308, opened 2023-02-20
Make `--[no-]show-...` options work with JSON output

pnax:
## Purpose

The CLI has several options to specify which field to display. Such options are not available when using JSON output `--json` or `--json-stream`. This PR make such options work with JSON output.

## Context

https://github.com/zonemaster/zonemaster-cli/issues/247#issuecomment-1436650786

## Changes

* Make `--[no-]show-...` options work with JSON output
* Display translated message tag when passing `--json` and `--json-translate`

## How to test this PR

Try several combinations using the `--[no-]show-...` and/or `--[no-]time` options and check that the fields are properly displayed or not. Also check that the combination `--json --json-translate` properly return the translated message.
```
$ zonemaster-cli --no-show-module --show-testcase --test connectivity --level INFO --json --json-translate afnic.fr | jq
[
  {
    "args": {
      "version": "v4.6.1"
    },
    "level": "INFO",
    "message": "Using version v4.6.1 of the Zonemaster engine.",
    "tag": "GLOBAL_VERSION"
  },
  {
    "args": {
      "asn_list": "2485,2486"
    },
    "level": "INFO",
    "message": "At least two IPv4 addresses of the authoritative nameservers are announce by different AS sets. A merged list of all AS: (2485,2486).",
    "tag": "IPV4_DIFFERENT_ASN"
  },
  {
    "args": {
      "asn_list": "2485,2486"
    },
    "level": "INFO",
    "message": "At least two IPv6 addresses of the authoritative nameservers are announce by different AS sets. A merged list of all AS: (2485,2486).",
    "tag": "IPV6_DIFFERENT_ASN"
  }
]
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -91,7 +91,7 @@ has 'json_translate' => (
     default       => 0,
     cmd_aliases   => 'json_translate',
     cmd_flag      => 'json-translate',
-    documentation => __( 'Flag indicating if streaming JSON output should include the translated message of the tag or not.' ),
+    documentation => __( 'Flag indicating if JSON output should include the translated message of the tag or not.' ),
 );
 
 has 'raw' => (
@@ -610,6 +610,11 @@ sub run {
         my $res = Zonemaster::Engine->logger->json( $self->level );
         $res = $JSON->decode( $res );
         foreach ( @$res ) {
+            if ( $self->json_translate ) {
+                my %e = %$_;
+                my $entry = Zonemaster::Engine::Logger::Entry->new( \%e );
+                $_->{message} = $json_translator->translate_tag( $entry );
+            }
             delete $_->{timestamp} unless $self->time;
             delete $_->{level} unless $self->show_level;
             delete $_->{module} unless $self->show_module;
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -430,11 +430,11 @@ sub run {
                 elsif ( $self->json_stream ) {
                     my %r;
 
-                    $r{timestamp} = $entry->timestamp;
-                    $r{module}    = $entry->module;
-                    $r{testcase}  = $entry->testcase;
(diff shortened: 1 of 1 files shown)


matsduf:
Also see https://github.com/zonemaster/zonemaster-cli/issues/247#issuecomment-1438411947

This is fine, but the documentation must be updated:
```
        --[no-]json-translate  Flag indicating if streaming JSON output
                               should include the translated message of the
                               tag or not.
```

What does `--raw` mean in combination with `--json`?

When `--raw` is given without `--json` then you get the tag and the arguments for that tag. Can you get the arguments here?

pnax:
> What does --raw mean in combination with --json?

`--raw` only applies to the message tag. `--json --raw` would mean output the message tag without translating it into any human language.

> When --raw is given without --json then you get the tag and the arguments for that tag. Can you get the arguments here?

I'm not sure I understand your question. Could you explain?

pnax:
I've updated the documentation for `--json-translate`.

matsduf:
> > When --raw is given without --json then you get the tag and the arguments for that tag. Can you get the arguments here?
> 
> I'm not sure I understand your question. Could you explain?

Forget my comment. Now it looks OK.

matsduf:
This looks fine, but could `--raw` get more complete documentation, e.g.
```
	--[no-]raw     Flag indicating if output should be translated
	                       to human language or dumped raw. This flag is
                               ignored if --json or --json-stream has been given.
```

matsduf:
>     * I'm planning in a follow-up PR to deprecate `--[no-]json-translate` and replace it with `--[no-]raw`.

So you mean that `--json-translate` will be replaced by `--no-raw`  and `--no-json-translate` by `--raw`?

If the end result is that it will be clear in the documentation what options that can be combined and not that I am fine with it.

pnax:
> So you mean that --json-translate will be replaced by --no-raw and --no-json-translate by --raw?

Yes. This is exactly what I intend to do.

matsduf (approved):
Requires additional PRs, but this part is fine.

PR #309, opened 2023-02-27
Improve implementation for JSON and raw options

pnax:
## Purpose

Continue the work to align the Zonemaster-CLI implementation with its own documentation.

## Context

Follow-up on #308 
https://github.com/zonemaster/zonemaster-cli/issues/247#issuecomment-1436650786

## Changes

* Make `--json-stream` entail `--json`
* Deprecate `--[no-]json-translate`, use `--[no-]raw` instead
* Refactor CLI.pm

## How to test this PR

Test several configurations mixing `--json`, `--raw`, `--json-stream`.
```
zonemaster-cli --json-stream --raw zonemaster.net
zonemaster-cli --json-stream zonemaster.net
zonemaster-cli --json zonemaster.net
zonemaster-cli --json --raw zonemaster.net
zonemaster-cli --json --json-stream --raw zonemaster.net
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -419,7 +419,7 @@ sub run {
             if ( $numeric{ uc $entry->level } >= $numeric{ $self->level } ) {
                 $printed_something = 1;
 
-                if ( $translator ) {
+                if ( not $self->raw and not $self->json ) {
                     my $header = q{};
                     if ( $self->time ) {
                         $header .= sprintf "%*.2f ", ${field_width{seconds}}, $entry->timestamp;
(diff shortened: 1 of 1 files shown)


marc-vanderwal:
A big improvement, definitely, which reduces the combinations of command-line options that do nothing (e.g. `--no-json-translate` without having given either `--json` or `--json-stream`).

With this design, all combinations make sense, except for one: `--json-stream --no-json`. Currently, it does the same thing as either `--json-stream` alone or `--json-stream --json`. I think `zonemaster-cli` should error out instead.

Because `--[no-]json-translate` is going to be deprecated, it may be a good thing to warn the user, and tell them the new set of options to use. For example:

- if `--no-json-translate` is given, tell them they should pass `--raw` instead;
- if `--json-translate` is given, tell them they should pass `--no-raw` instead;
- additionally, if `--no-json-translate` or `--json-translate` is given without either `--json` or `--json-stream`, tell the user that the option has no effect without either of those two.

pnax:
Thanks for the review and comments. I've updated the PR to display error and/or warning messages to the user.

pnax:
Rebased to fix conflicts and incorporate latest release changes. Please re-review.

marc-vanderwal (approved):
Looks good to me.

PR #314, opened 2023-03-03
Unit test for PO files

pnax:
## Purpose

New unit test to verify that PO files are correct regarding some rules:
* no fuzzy marks
* all PO files are tidied
* PO files msgid/msgstr format should be consistent 

## Context

https://github.com/zonemaster/zonemaster-cli/issues/277

## Changes

new t/po-files.t

## How to test this PR

Unit test should pass.

### DIFF
--- a/share/GNUmakefile
+++ b/share/GNUmakefile
@@ -1,6 +1,6 @@
 .POSIX:
 .SUFFIXES: .po .mo
-.PHONY: all check-msg-args dist extract-pot tidy-po show-fuzzy update-po new-po
+.PHONY: all check-msg-args dist extract-pot tidy-po show-fuzzy update-po new-po check-po
 
 POFILES := $(shell find . -maxdepth 1 -type f -name '*.po' -exec basename {} \;)
 MOFILES := $(POFILES:%.po=%.mo)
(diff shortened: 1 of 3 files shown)


share/GNUmakefile
```diff
@@ -29,6 +29,9 @@ new-po: extract-pot
 		-e 's/^("Content-Type:.+charset=)CHARSET/$${1}UTF-8/;' $(POLANG).po
 	@perl -ni -e 'print unless /^#( |$$)/' $(POLANG).po
 
+check-po:
+	@for f in $(POFILES) ; do msgfmt -c $$f ; done
+
```
matsduf, line 34: Please add a comment that describes what check-po does. Just 1-2 sentences.

matsduf (commented):
The Zonemaster-Engine /util/check-msg-args is not available here. Currently there are only a few arguments, but that might change.

I suggest that we have the script in all three environments where we have PO files, Engine, CLI and Backend, to make it consistent and the instructions can be the same.

I also suggest that check-msg-args script is put in the share directory to make it easier for the translator to execute. It is really only meaningful for the translation work. (Moving/putting it into share should apply for all three environments.)

The same comment has been added to https://github.com/zonemaster/zonemaster-backend/pull/1089

pnax:
> the untidy PO file isn’t caught.

Did you commit the change? In order to test PO file tidiness, the git repository needs to be in a clean state (nothing in the staging area).

marc-vanderwal:
Oh okay. After I made a commit breaking the formatting of one of the PO files, the untidy PO file is caught indeed.

pnax:
I could agree that this is maybe a quick and hacky way of checking tidiness.

share/GNUmakefile
matsduf: My suggestion:
```
# Check the msgid/msgstr pair for some inconsistencies between them in the selected PO file and
# report on standard error any errors found. The PO file is not updated.
```

marc-vanderwal:
> I could agree that this is maybe a quick and hacky way of checking tidiness.

There doesn’t seem to be a good clean way to check this. At least it’ll get caught in continuous integration tests.

matsduf (approved):

pnax:
> I suggest that we have the script in all three environments where we have PO files, Engine, CLI and Backend, to make it consistent and the instructions can be the same.

This could be a good idea. Not sure in which folder the script would fit better though. Could you do it in a follow-up PR?

matsduf:
> This could be a good idea. Not sure in which folder the script would fit better though. Could you do it in a follow-up PR?

Sure.

matsduf (approved):

Issue #315, opened 2023-02-09
sourceaddr for IPv6 and IPv4


simtomas:
Hi,

while it is possible to specify --sourceaddr with either IPv6 or IPv4, 
it is not possible to specify both.
Thus if v4 is specified all v6 tests fail (cannot contact servers via v6) and vice versa.

Would it be possible to add second parameter like --sourceaddr6, to specify v6 address in addition to 
--sourceaddr v4 address?

matsduf:
This issue requires changes to both Zonemaster::Engine (see https://github.com/zonemaster/zonemaster-engine/pull/1203) and Zonemaster::CLI.

pnax:
> Would it be possible to add second parameter like --sourceaddr6, to specify v6 address in addition to
--sourceaddr v4 address?

Two new options `--sourceaddr4` and `--sourceaddr6` have been added (#1203) and will be available with v2023.1. I'm closing this issue.

Issue #316, opened 2023-03-13
Incorrect IP address with --sourceaddr gives not error message


matsduf:
The documentation for `--sourceaddr` says
```
	--sourceaddr STR       Source IP address used to send queries.
	                       Setting an IP address not correctly configured
	                       on a local network interface causes cryptic
	                       error messages.
```
It is expected that some error message comes out if the IP address is wrong (not configured on the interface), but when `--sourceaddr` is used in that way nothing comes out that gives a clue.

Compare a test with correct IP address with a one that is not on the interface (IPv6 disabled):
```
$ zonemaster-cli --test basic --level info zonemaster.net --no-ipv6 --sourceaddr 172.31.10.109 # Correct IP address
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.6.2 of the Zonemaster engine.
   0.28 INFO      Parent domain 'net' was found for the tested domain.
   1.28 INFO      Nameserver ns2.nic.fr/192.93.0.4 listed these servers as glue: ns2.nic.fr.,nsa.dnsnode.net.,nsp.dnsnode.net.,nsu.dnsnode.net..
   1.28 INFO      Nameserver nsa.dnsnode.net/194.58.192.46 listed these servers as glue: ns2.nic.fr.,nsa.dnsnode.net.,nsp.dnsnode.net.,nsu.dnsnode.net..
   1.28 INFO      Nameserver nsp.dnsnode.net/194.58.198.32 listed these servers as glue: ns2.nic.fr.,nsa.dnsnode.net.,nsp.dnsnode.net.,nsu.dnsnode.net..
   1.29 INFO      Nameserver nsu.dnsnode.net/185.42.137.98 listed these servers as glue: ns2.nic.fr.,nsa.dnsnode.net.,nsp.dnsnode.net.,nsu.dnsnode.net..
   1.29 INFO      Functional nameserver found. "A" query for www.zonemaster.net test skipped.

$ zonemaster-cli --test basic --level info zonemaster.net --no-ipv6 --sourceaddr 172.31.10.110 # Incorrect IP address
Seconds Level     Message
======= ========= =======
   0.00 INFO      Using version v4.6.2 of the Zonemaster engine.
   0.02 INFO      Parent domain '.' was found for the tested domain.
   0.04 CRITICAL  No NS records for tested zone from parent. NS tests skipped.
```
Preferably something should come out that indicate the error.

PR #317, opened 2023-03-13
New options "--sourceaddr4" and "--sourceaddr6"

pnax:
## Purpose

Provide options to define both source IPv4 and source IPv6. This would replace the `--sourceaddr` option.

## Context

#315 
Follow-up on https://github.com/zonemaster/zonemaster-engine/pull/1203

## Changes

* New options "--sourceaddr4" and "--sourceaddr6"
* Deprecate "--sourceaddr".

## How to test this PR

```
zonemaster-cli --sourceaddr <IP address> bzh # should emit a deprecation message
zonemaster-cli --sourceaddr4 <IPv4 address> berlin
zonemaster-cli --sourceaddr6 <IPv6 address> tokyo
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -278,8 +278,32 @@ has 'sourceaddr' => (
     isa           => 'Str',
     required      => 0,
     documentation => __(
-            'Source IP address used to send queries. '
+            'Deprecated (planned removal: v2024.1). '
+          . 'Use --sourceaddr4 and/or --sourceaddr6. '
+          . 'Source IP address used to send queries. '
           . 'Setting an IP address not correctly configured on a local network interface causes cryptic error messages.'
+    ),
+);
+
+has 'sourceaddr4' => (
+    is            => 'ro',
+    isa           => 'Str',
+    required      => 0,
+    documentation => __(
+            'Source IPv4 address used to send queries. '
+          . 'Setting an IPv4 address not correctly configured on a local network interface fails silently. '
+          . 'Can not be combined with --sourceaddr.'
+    ),
+);
+
+has 'sourceaddr6' => (
+    is            => 'ro',
+    isa           => 'Str',
+    required      => 0,
+    documentation => __(
+            'Source IPv6 address used to send queries. '
+          . 'Setting an IPv6 address not correctly configured on a local network interface fails silently. '
+          . 'Can not be combined with --sourceaddr.'
     ),
 );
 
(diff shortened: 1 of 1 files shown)


lib/Zonemaster/CLI.pm
```diff
@@ -278,11 +278,32 @@ has 'sourceaddr' => (
...
     ),
 );
 
+has 'sourceaddr4' => (
+    is            => 'ro',
+    isa           => 'Str',
+    required      => 0,
+    documentation => __(
+            'Source IPv4 address used to send queries. '
+          . 'Setting an IPv4 address not correctly configured on a local network interface fails silently.'
```
matsduf, line 293: Add 'Setting an IP address not correctly configured on a local network interface causes cryptic error messages.' and "Cannot be combined with --sourceaddr".

lib/Zonemaster/CLI.pm
```diff
@@ -278,11 +278,32 @@ has 'sourceaddr' => (
...
+    ),
+);
+
+has 'sourceaddr6' => (
+    is            => 'ro',
+    isa           => 'Str',
+    required      => 0,
+    documentation => __(
+            'Source IPv6 address used to send queries. '
+          . 'Setting an IPv6 address not correctly configured on a local network interface fails silently.'
```
matsduf, line 303: Add 'Setting an IP address not correctly configured on a local network interface causes cryptic error messages.' and "Cannot be combined with --sourceaddr".

matsduf (changes requested):
Add a blocker so that --sourceaddr can neither be combined with --sourceaddr4 nor --sourceaddr6.

lib/Zonemaster/CLI.pm
pnax: > Add 'Setting an IP address not correctly configured on a local network interface causes cryptic error messages.'

From #316 I understand that there won't be such error messages. Do I understand it correctly?

pnax:
> 

Add a blocker so that --sourceaddr can neither be combined with --sourceaddr4 nor --sourceaddr6.

lib/Zonemaster/CLI.pm
matsduf: True, but I still think we should have the warning. Suggestion:
```
Setting an IP address not correctly configured on a local network interface will silently
prevent Zonemaster-CLI from working correctly.
```

matsduf:
It should be noted that `zonemaster-cli` will this update will require Zonemaster::Engine updated as updated with https://github.com/zonemaster/zonemaster-engine/pull/1203.

matsduf (dismissed):

lib/Zonemaster/CLI.pm
matsduf: Yes. And your text is fine.

matsduf (approved):

Issue #248, opened 2022-01-07
Expand the json format


matsduf:
If --json option is selected then only the result from the test case tests are included in the JSON structure. The data that --nstimes and --elapsed are printed outside the JSON structure in non-JSON format.

Expand the JSON format to be able to include also the data from --nstimes and --elapsed.

Also see #245.

PR #318, opened 2023-03-14
--nstimes and --elapsed in JSON format

pnax:
## Purpose

Expand the JSON format to include the data from --nstimes and --elapsed.

## Context

#248

## Changes

* output times in JSON format

## How to test this PR

```
zonemaster-cli --json --nstimes madrid
zonemaster-cli --json --elapsed london
```


lib/Zonemaster/CLI.pm
```diff
@@ -622,24 +622,49 @@ sub run {
...
+                $times{$ns} = {
+                    'max'    => sprintf( '%.2f', 1000 * $ns->max_time ),
+                    'min'    => sprintf( '%.2f', 1000 * $ns->min_time ),
+                    'avg'    => sprintf( '%.2f', 1000 * $ns->average_time ),
+                    'stddev' => sprintf( '%.2f', 1000 * $ns->stddev_time ),
+                    'median' => sprintf( '%.2f', 1000 * $ns->median_time ),
+                    'total'  => sprintf( '%.2f', 1000 * $ns->sum_time )
+                };
+            }
+            say $JSON->encode( \%times );
```
marc-vanderwal, line 637: By using `sprintf`, the times in the JSON object are strings. Why not floating-point numbers?

lib/Zonemaster/CLI.pm
```diff
@@ -622,24 +622,49 @@ sub run {
...
         }
     }
 
     if ($self->elapsed) {
         my $last = Zonemaster::Engine->logger->entries->[-1];
-        printf "Total test run time: %0.1f seconds.\n", $last->timestamp;
+
+        if ( $self->json ) {
+            my %elapsed = (
+                'elapsed' => sprintf( '%.2f', $last->timestamp )
```
marc-vanderwal, line 661: The `%.2f` format string is inconsistent with the one in use when JSON output is disabled. And why represent that timestamp as a string instead of a floating-point number in the JSON object?

pnax: An error. I've removed the rounding (see previous comment).

matsduf:
@pnax, can you rebase this on update develop where the PO file tests are removed from distribution package?

matsduf:
`zonemaster-cli --json` will output the result in one JSON structure, but with `--nstimes` or `--elapsed` there will be two structures (or three if both are included). Why not include the elapsed and nstimes structure into the same structure?

If `json-stream` is selected instead of `json` everything is presented on the same, top, level.

pnax:
> zonemaster-cli --json will output the result in one JSON structure [...] Why not include the elapsed and nstimes structure into the same structure?

This structure is a JSON array composed of similar items. Adding the output of --nselapsed or --nstimes to this array would result in an item inconsistency.

matsduf:
> > zonemaster-cli --json will output the result in one JSON structure [...] Why not include the elapsed and nstimes structure into the same structure?
> 
> This structure is a JSON array composed of similar items. Adding the output of --nselapsed or --nstimes to this array would result in an item inconsistency.

One option would be to create a new structure to be used is `elapsed` or `nstimes` is used where the array of result objects is named:
```JSON
{
  "elapsed": {
      "elapsed": "0.54"
    },
  "nstimes": [
     {...},
     {...}
  ],
  "jobresults": [
     {...},
     {...}
  ]
}
```

Also the documentation should be updated for `--elapsed` and `--nstimes`

```
$ zonemaster-cli -h
(...)
	--[no-]nstimes         At the end of a run, print a summary of the
	                       times (in milliseconds) the zone's name
	                       servers took to answer.
(...)
	--[no-]elapsed         Print elapsed time (in seconds) at end of run.
```

When running with `--json` the appear first.

pnax:
We could provide a different structure indeed. From your suggestion, I suggest something a little bit different for the `nstimes` array: 
```json
{
  "elapsed": {
      "elapsed": 0.54
    },
  "nstimes": [
   {
      "avg": 3.39,
      "max": 3.39,
      "median": 3.39,
      "min": 3.39,
      "server/ip": "nsa.dnsnode.net/194.58.192.46",
      "stddev": 0.00,
      "total": 3.39
    },
    {...}
  ],
  "jobresults": [
    {
      "args": {
        "version": "v4.6.2"
      },
      "level": "INFO",
      "message": "Using version v4.6.2 of the Zonemaster engine.",
      "tag": "GLOBAL_VERSION",
      "testcase": "UNSPECIFIED",
      "timestamp": 0.000745058059692383
     },
     {...}
  ]
}
```

matsduf:
> We could provide a different structure indeed. From your suggestion, I suggest something a little bit different for the `nstimes` array:

I think your suggestion is fine. 

Is `server/ip` a good key name? I suggest `nameserver` or even rather `ns`. In messages from test cases the standardized argument name for nameserver in the form "NAME/IP" is `ns` and I think it makes sense to use the same.

pnax:
A new proposal based on your comment and adding the output of the `--count` option:

```json
{
  "count": {                                                                                                                            
    "DEBUG": 769,                          
    "DEBUG2": 2306,                                                                                                                     
    "DEBUG3": 1727,                                                                                                                     
    "INFO": 97,                                                                                                                         
    "NOTICE": 1,                                                                                                                        
    "WARNING": 2                                                                                                                        
  },
  "elapsed": 0.54,
  "nstimes": [
   {
      "avg": 3.39,
      "max": 3.39,
      "median": 3.39,
      "min": 3.39,
      "ns": "nsa.dnsnode.net/194.58.192.46",
      "stddev": 0.00,
      "total": 3.39
    },
    {...}
  ],
  "jobresults": [
    {
      "args": {
        "version": "v4.6.2"
      },
      "level": "INFO",
      "message": "Using version v4.6.2 of the Zonemaster engine.",
      "tag": "GLOBAL_VERSION",
      "testcase": "UNSPECIFIED",
      "timestamp": 0.000745058059692383
     },
     {...}
  ]
}
```

lib/Zonemaster/CLI.pm
```diff
@@ -259,7 +259,7 @@ has 'nstimes' => (
     isa           => 'Bool',
     required      => 0,
     default       => 0,
-    documentation => __('At the end of a run, print a summary of the times the zone\'s name servers took to answer.'),
+    documentation => __( 'At the end of a run, print a summary of the times (in milliseconds) the zone\'s name servers took to answer.' ),
```
matsduf, line 262: This could be interpreted that it is presented at the end, which is not necessarily true for JSON output.

lib/Zonemaster/CLI.pm
```diff
@@ -312,7 +312,7 @@ has 'elapsed' => (
     isa           => 'Bool',
     required      => 0,
     default       => 0,
-    documentation => 'Print elapsed time at end of run.',
+    documentation => __( 'Print elapsed time (in seconds) at end of run.' ),
```
matsduf, line 315: This could be interpreted that it is presented at the end, which is not necessarily true for JSON output.

matsduf (approved):
Two minor comments, but this looks fine and I approve.

matsduf:
@marc-vanderwal, you had some comments. Please re-review.

marc-vanderwal (approved):
Looks good to me.

PR #333, opened 2023-05-25
Extend --test option to allow passing only a testcase

### LINKED ISSUE #232: Update the --test option
If only the test cases in a module or a single test case is to be run, then the `--test` option can be used:
```
	--test STR...          Specify test to run. Should be either the name
	                       of a module, or the name of a module and the
	                       name of a method in that module separated by a
	                       "/" character (Example: "Basic/basic1"). The
	                       method specified must be one that takes a zone
	                       object as its single argument. This switch can
	                       be repeated.
```

To run all Basic test cases run `zonemaster-cli --test basic`, but to test only test case Basic04, then "basic" has to be repeated, `zonemaster-cli --test basic/basic04`. The module can always be deduced from the test case identifier, and the difference between test case identifier vs. module name can always be deduced from the existence of digits a the end.

This issue suggests that the value of the `--test` option should also accept the test case identifier only. I.e. all the following three forms:

* --test basic # test all test cases in the basic module
* --test basic/basic04 # test Basic04 only
* --test basic04 # test Basic04 only

pnax:
## Purpose

Make it possible to only pass the name of a testcase to the `--test` option.

The following commands are now both ok and identical:
```
zonemaster-cli --test basic/basic01
zonemaster-cli --test basic01
```

## Context

Fixes #232

## Changes

* allow passing only a testcase name to the `--test` option

## How to test this PR

* run all testcases from the Basic module:
   ```
   zonemaster-cli --test basic
   ```
* run only one testcase
   ```
   zonemaster-cli --test basic/basic01
   zonemaster-cli --test basic01
   ```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -192,7 +192,7 @@ has 'test' => (
     isa           => 'ArrayRef',
     required      => 0,
     documentation => __(
-'Specify test to run case-insensitively. Should be either the name of a module, or the name of a module and the name of a method in that module separated by a "/" character (Example: "Basic/basic01"). This switch can be repeated.'
+'Specify one or several test cases to be run. Should be the case-insensitive name of a test module and/or a test case (examples: "Delegation", "delegation01", "Delegation/delegation01"). This switch can be repeated.'
     )
 );
 
@@ -621,13 +621,24 @@ sub run {
     # Actually run tests!
     eval {
         if ( $self->test and @{ $self->test } > 0 ) {
+            my $zone = Zonemaster::Engine->zone( $domain );
             foreach my $t ( @{ $self->test } ) {
-                my ( $module, $method ) = split( '/', lc($t), 2 );
-                if ( $method ) {
-                    Zonemaster::Engine->test_method( $module, $method, Zonemaster::Engine->zone( $domain ) );
+                # The case does not matter
+                $t = lc( $t );
+                # Fully qualified module and test case (e.g. Example/example12)
+                if (my ($module, $method) = $t =~ m#^ ( [a-z]+ ) / ( [a-z]+[0-9]{2} ) $#ix) {
+                    Zonemaster::Engine->test_method( $module, $method, $zone );
                 }
+                # Just a test case (e.g. example12). Note the different capturing order.
+                elsif (($method, $module) = $t =~ m#^ ( ( [a-z]+ ) [0-9]{2} ) $#ix) {
+                    Zonemaster::Engine->test_method( $module, $method, $zone );
+                }
+                # Just a module name (e.g. Example) or something invalid.
+                # TODO: in case of invalid input, print a proper error message
+                # suggesting to use --list-tests for valid choices.
                 else {
-                    Zonemaster::Engine->test_module( $module, $domain );
+                    $t =~ s/\/$//;
+                    Zonemaster::Engine->test_module( $t, $zone );
                 }
             }
         }
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -192,7 +192,7 @@ has 'test' => (
     isa           => 'ArrayRef',
     required      => 0,
     documentation => __(
-'Specify one or several test cases to be run. Should be the case-insensitive name of a test module and/or a test case (examples: "Delegation", "delegation01", "Delegation/delegation01"). This switch can be repeated.'
+'Specify test case to be run. Should be the case-insensitive name of a test module (e.g. "Delegation") and/or a test case (e.g. "Delegation/delegation01" or "delegation01"). This switch can be repeated.'
     )
 );
 
@@ -637,8 +637,8 @@ sub run {
                 # TODO: in case of invalid input, print a proper error message
                 # suggesting to use --list-tests for valid choices.
                 else {
-                    $t =~ s/\/$//;
-                    Zonemaster::Engine->test_module( $t, $zone );
+                    $t =~ s{/$}{};
+                    Zonemaster::Engine->test_module( $t, $domain );
                 }
             }
         }
--- a/script/zonemaster-cli
+++ b/script/zonemaster-cli
@@ -196,12 +196,14 @@ Default: on
 
 Print all test cases listed in the test modules, then exit.
 
-=item --test=MODULE, --test=MODULE/TESTCASE
+=item --test=MODULE, --test=MODULE/TESTCASE, --test=TESTCASE
 
 Limit the testing suite to run only the specified tests.
 This can be the name of a testing module, in which case all test cases from
 that module will be run, or the name of a module followed by a slash and the
-name of a test case (test case identifier) in that module.
+name of a test case (test case identifier) in that module, or the name of the
+test case.
+Can be specified multiple times.
 This option is case-insensitive.
 
 =item --stop_level=LEVEL, --stop-level=LEVEL


lib/Zonemaster/CLI.pm
```diff
@@ -618,6 +618,10 @@ sub run {
         if ( $self->test and @{ $self->test } > 0 ) {
             foreach my $t ( @{ $self->test } ) {
                 my ( $module, $method ) = split( '/', $t, 2 );
+                if ( $module =~ /^([A-Za-z_]+)[0-9]+$/ ) {
+                    $method = $module;
+                    $module = $1;
+                }
                 if ( $method ) {
                     Zonemaster::Engine->test_method( $module, $method, Zonemaster::Engine->zone( $domain ) );
                 }
```
matsduf, line 627: I think it would be clearer if $method is $testcase.

pnax: I thought about that too. There is a logic to use `method` since this is the name of the implemented method in the code. I don't have a strong opinion on which word to use.

lib/Zonemaster/CLI.pm
```diff
@@ -194,7 +194,7 @@ has 'test' => (
     isa           => 'ArrayRef',
     required      => 0,
     documentation => __(
-'Specify test to run. Should be either the name of a module, or the name of a module and the name of a method in that module separated by a "/" character (Example: "Basic/basic1"). The method specified must be one that takes a zone object as its single argument. This switch can be repeated.'
+'Specify test to run. Should be either the name of a module, the name of a module and the name of a method in that module separated by a "/" character (example: "Basic/basic01") or the name of a method. The method specified must be one that takes a zone object as its single argument. This switch can be repeated.'
```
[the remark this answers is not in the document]
pnax: > Is there any good reason, besides backward compatibility, to keep the basic/basic01 format?

I think backward compatibility is a good reason enough.

> The way "test" is implemented, if you specify more than one test case Zonemaster is run once per test case. Do you see that it would be possible to run all such test cases in one session? That would be an improvement.

I've looked a little bit into it, that would require some work in Zonemaster-Engine and Zonemaster-CLI. If you have an idea of the expected design, can you open an issue with it?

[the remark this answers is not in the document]
matsduf: The refer to the test cases as "test case" in most cases against the users. For the users it is not relevant if the test cases are implemented as methods in the code or not.

lib/Zonemaster/CLI.pm
matsduf: There are many methods. $testcase makes it easier to read the code.

lib/Zonemaster/CLI.pm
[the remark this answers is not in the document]
pnax: > In most cases test cases are referred to as "test case" for the users. For the users it is not relevant if the test cases are implemented as methods in the code or not. The documentation should better be updated.

I'm not sure I understand what you mean here. Could you explain please?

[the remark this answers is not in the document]
matsduf: In
```
'Specify test to run. Should be either the name of a module, the name of a module and the name of a method in that module separated by a "/" character (example: "Basic/basic01") or the name of a method. The method specified must be one that takes a zone object as its single argument. This switch can be repeated.'
```
please update `the name of a method` (two places) to `the name of a test case` to be consistent with other documents.

What does `The method specified must be one that takes a zone object as its single argument` mean? Are there any test cases that do not need a single zone name? Does the user ever provide a `zone object`? Isn't that an internal unit?

lib/Zonemaster/CLI.pm
matsduf: Todays code permits `basic/basic01` and `BASIC/basic01`, but it does not permit `basic/BASIC01`. I see that as a bug. Could that be fixed in this PR?

lib/Zonemaster/CLI.pm
```diff
@@ -623,6 +623,10 @@ sub run {
         if ( $self->test and @{ $self->test } > 0 ) {
             foreach my $t ( @{ $self->test } ) {
                 my ( $module, $method ) = split( '/', $t, 2 );
+                if ( $module =~ /^([A-Za-z_]+)[0-9]+$/ ) {
```
matsduf, line 626: Why is underscore "_" included in the module name part? There is no such case and no plan for that.

pnax: I think it's because I couldn't find any place defining the character set for the module name. Maybe I overlooked something.

lib/Zonemaster/CLI.pm
pnax: Maybe this could be done in an independent PR, see #343.

lib/Zonemaster/CLI.pm
matsduf: In https://github.com/zonemaster/zonemaster/blob/master/docs/internal/templates/specifications/tests/TestCaseIdentifierSpecification.md the test levels (modules) are defined and all matches /^[A-Z]+$/i.

pnax: Updated

lib/Zonemaster/CLI.pm
```diff
@@ -192,7 +192,7 @@ has 'test' => (
     isa           => 'ArrayRef',
     required      => 0,
     documentation => __(
-'Specify test to run case-insensitively. Should be either the name of a module, or the name of a module and the name of a method in that module separated by a "/" character (Example: "Basic/basic01"). This switch can be repeated.'
+'Specify test to run case-insensitively. Should be either the name of a module, the name of a module and the name of a testcase in that module separated by a "/" character (example: "Delegation/delegation01") or the name of a testcase. This switch can be repeated.'
```
matsduf, line 195: In other cases in this file, it is written "test case". I think it should be consistent.

lib/Zonemaster/CLI.pm
```diff
@@ -623,6 +623,10 @@ sub run {
         if ( $self->test and @{ $self->test } > 0 ) {
             foreach my $t ( @{ $self->test } ) {
                 my ( $module, $method ) = split( '/', lc($t), 2 );
+                if ( $module =~ /^([a-z]+)[0-9]+$/ ) {
+                    $method = $module;
+                    $module = $1;
+                }
```
matsduf, line 629: If $module matches a test case identifier then $method can be any garbage. I guess that is OK in this context.

matsduf (approved):

lib/Zonemaster/CLI.pm
```diff
@@ -192,7 +192,7 @@ has 'test' => (
     isa           => 'ArrayRef',
     required      => 0,
     documentation => __(
-'Specify test to run case-insensitively. Should be either the name of a module, or the name of a module and the name of a method in that module separated by a "/" character (Example: "Basic/basic01"). This switch can be repeated.'
+'Specify test to run case-insensitively. Should be either the name of a module, the name of a module and the name of a test case in that module separated by a "/" character (example: "Delegation/delegation01") or the name of a test case. This switch can be repeated.'
```
mattias-p, line 195: I'm not particularly fond of this description so I tried to improve it. Here's my take:

```suggestion
'Specify test to run. Should be the case-insensitive name of a test module and/or a test case. (Examples: "Delegation", "delegation01", "Delegation/delegation01"). This switch can be repeated.'
```

lib/Zonemaster/CLI.pm
```diff
@@ -623,6 +623,10 @@ sub run {
         if ( $self->test and @{ $self->test } > 0 ) {
             foreach my $t ( @{ $self->test } ) {
                 my ( $module, $method ) = split( '/', lc($t), 2 );
+                if ( $module =~ /^([a-z]+)[0-9]+$/ ) {
```
mattias-p, line 626: Shouldn't this regex use the `i` flag?

matsduf: Note the `lc` in `lc($t)`.

matsduf: But to prevent "basic01/garbage" to be taken as "basic/basic01", I suggest
```
                my ( $module, $method ) = split( '/', lc($t), 2 );
                if ( $module =~ /^([a-z]+)[0-9]+$/  and !$method) {
```

lib/Zonemaster/CLI.pm
matsduf: "Specify one or several test cases to be run."

lib/Zonemaster/CLI.pm
mattias-p: We should also forbid "basic01/" and "basic01/0".

marc-vanderwal (commented):
Good idea, but I think there is room for improvement.

lib/Zonemaster/CLI.pm
pnax: Good idea. Currently the message comes from the Engine which has no idea of the `--list-tests` CLI option. I guess it would mean rethinking how this message/error is emitted.

> But that’s a problem best addressed in another PR.

I'll open an issue with your suggestion.

lib/Zonemaster/CLI.pm
pnax: Updated with both suggestions.

matsduf (approved):
I think this is a great improvement.

marc-vanderwal:
I'd approve it, but it’s best if someone else had a look at the code.

lib/Zonemaster/CLI.pm
```diff
@@ -621,13 +621,24 @@ sub run {
...
+                # Just a test case (e.g. example12). Note the different capturing order.
+                elsif (($method, $module) = $t =~ m#^ ( ( [a-z]+ ) [0-9]{2} ) $#ix) {
+                    Zonemaster::Engine->test_method( $module, $method, $zone );
+                }
+                # Just a module name (e.g. Example) or something invalid.
+                # TODO: in case of invalid input, print a proper error message
+                # suggesting to use --list-tests for valid choices.
                 else {
-                    Zonemaster::Engine->test_module( $module, $domain );
+                    $t =~ s/\/$//;
```
mattias-p, line 640: Just a tiny suggestion on readability:
```suggestion
                    $t =~ s{/$}{};
```

script/zonemaster-cli
```diff
@@ -196,12 +196,14 @@ Default: on
...
-=item --test=MODULE, --test=MODULE/TESTCASE
+=item --test=MODULE, --test=MODULE/TESTCASE, --test=TESTCASE
 
 Limit the testing suite to run only the specified tests.
 This can be the name of a testing module, in which case all test cases from
 that module will be run, or the name of a module followed by a slash and the
-name of a test case (test case identifier) in that module.
+name of a test case (test case identifier) in that module, or the name of the
+test case.
+Can be specified multiple times.
```
mattias-p, line 206: > Can be specified multiple times.

I propose that we postpone this clarification until #359. That PR (at least in its current state) makes repeated `--test` arguments overwrite each other. So if we include this PR in the upcoming release and #359 in the next one, we can avoid some flip-flopping in our documentation.

mattias-p:
### v2023.2 Release Testing

I successfully tested this on Rocky Linux 8.9.

PR #337, opened 2023-05-31
Fix option --[no-]json-translate

pnax:
## Purpose

This PR fixes 2 small error/bugs discovered during v2023.1 release testing.

## Context

#308 https://github.com/zonemaster/zonemaster-cli/pull/309
v2023.1 release testing comment https://github.com/zonemaster/zonemaster-cli/pull/308#issuecomment-1570279972
v2023.1 release testing comment https://github.com/zonemaster/zonemaster-cli/pull/309#issuecomment-1570297400

## Changes

* fix ineffective `--no-json-translate` option
* show deprectation warning with option alias `--[no-]json_translate`

## How to test this PR

* Check that `--[no-]json_translate` option gives a deprecation warning. 
* Check that `--no-json-translate` does not output the `message` key.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -88,7 +88,6 @@ has 'json_translate' => (
     traits        => [ 'Getopt' ],
     is            => 'ro',
     isa           => 'Bool',
-    default       => 0,
     cmd_aliases   => 'json_translate',
     cmd_flag      => 'json-translate',
     documentation => __( 'Deprecated. Flag indicating if JSON output should include the translated message of the tag or not.' ),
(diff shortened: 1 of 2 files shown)


lib/Zonemaster/CLI.pm
```diff
@@ -390,6 +390,7 @@ sub run {
     # align values
     $self->json( 1 ) if $self->json_stream;
     $self->raw( 0 ) if $self->json_translate; # deprecated
+    $self->raw( 1 ) if ( not $self->json_translate ); # deprecated
```
mattias-p, line 393: Since we're always overriding `$self->raw` here, the `--raw` option is always ignored, right? It's probably better to only override `$self->raw` if `--[no-]translate-json` is actually specified.

pnax: > Since we're always overriding $self->raw here, the --raw option is always ignored, right?

Only if you specify both options, something like `--no-json-translate --no-raw` which is an unexpected (and unsupported) behavior.

pnax: Rereading your comment, I think I misunderstood it the first time. I know understand what you meant. `$self->json_translate` is set to `0` by default. So this condition has a flaw. Now that I see it, I'll provide an update.

lib/Zonemaster/CLI.pm
```diff
@@ -389,7 +389,10 @@ sub run {
 
     # align values
     $self->json( 1 ) if $self->json_stream;
-    $self->raw( 0 ) if $self->json_translate; # deprecated
+    if ( not grep {/^--raw$/ || /^--no-raw$/} @{$self->ARGV} ) {
```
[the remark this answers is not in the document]
pnax: With your comment I discovered that `--noraw` is a valid form for the option. From [Getopt::Long documentation](https://metacpan.org/pod/Getopt::Long#!) there are indeed 2 valid prefix `no` and `no-`.

I really like your suggestion. I'll refactor and use that instead.

pnax:
I've refactored the PR. There was a small glitch with your suggestion @mattias-p. If there is no option passed, the output should be translated (`raw == 0`). I've added some logic to handle this case as well.

mattias-p (approved):

PR #348, opened 2023-08-03
Preparation for release v2023.1.3 (CLI)

matsduf:
## Purpose

Preparation for release.

## Changes

Updates files `Changes`, `Makefile.PL` and `CLI.pm`.

## How to test this PR

Review the changes.

### DIFF
--- a/Changes
+++ b/Changes
@@ -1,5 +1,13 @@
 Release history for Zonemaster component Zonemaster-CLI 
 
+
+v6.0.2 2023-08-07 (public fix version)
+
+ [Fixes]
+ - This version contains no real changes. It has been created to
+   require a higher (fixed) version of Zonemaster-Engine.
+
+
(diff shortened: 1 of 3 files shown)


matsduf:
> Doesn't this PR will have to be merged no matter what, just because the required Engine version has changed? If so then I don't see how it relates to the issue mentioned with Docker.

No, not if we follow the method proposed in https://github.com/zonemaster/zonemaster/pull/1194. Note that the version in Makefile.PL is the minimum version, and `cpanm` will take the highest version available.

matsduf:
I suggest that we do not merge this and instead use the method proposed in https://github.com/zonemaster/zonemaster/pull/1194.

pnax:
>I suggest that we do not merge this

I do think it would make sense to update the CLI and Backend version during this fix release. The introduced bug in Zonemaster-Engine is critical enough to force the minimal required version in the other modules.

PR #357, opened 2023-10-18
Use Zonemaster::Engine::Normalization module for input name normalization

tgreenx:
## Purpose

This PR introduces the usage of the [Zonemaster::Engine::Normalization](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.3/lib/Zonemaster/Engine/Normalization.pm) module for input DNS name normalization.
It also adds input validation to the `--ns` argument..

## Context

The current specification can found [here](https://github.com/zonemaster/zonemaster/blob/v2023.1.4/docs/public/specifications/tests/RequirementsAndNormalizationOfDomainNames.md), and will replace the [Basic00](https://github.com/zonemaster/zonemaster/blob/v2023.1.4/docs/public/specifications/tests/Basic-TP/basic00.md) Test Case.

## Changes

- Make use of `Zonemaster::Engine::Normalization::normalize_name()`
- Significantly improve `--ns` argument validation:
   - Add check for several slash (`/`) characters
   - Add IP address validation which leverages both regex checks and `Net::IP::XS::Error()` for detailed error messages 
- Remove `Zonemaster::CLI::to_idn()`
- List `Readonly` and `Net::IP::XS` as direct dependencies

## How to test this PR

Testing should proceed as before, although more errors are now supported and several can be reported simultaneously:

```
$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 '@.İabcİ.İcc'
Ambiguous downcaseing of character "LATIN CAPITAL LETTER I WITH DOT ABOVE" in the domain name. Use all lower case instead.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01
Must give the name of a domain to test.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 ''
Must give the name of a domain to test.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 ' '
Domain name is empty.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 ' @!.'
Domain name has an ASCII label ("@!") with a character not permitted.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1/.zonemaster.net/10.0.0.1'
--ns must be a name or a name/ip pair.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net/1'
Invalid IP address in --ns argument.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net/10.0.0.Z'
Invalid IP address in --ns argument:
        Invalid chars in IP 10.0.0.Z
        
$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net/fe10:::'
Invalid IP address in --ns argument:
        Invalid address fe10::: (More than one :: pattern)

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1@.zonemaster@.net!.'
Invalid name in --ns argument:
        Domain name has an ASCII label ("ns1@") with a character not permitted.
        Domain name has an ASCII label ("zonemaster@") with a character not permitted.
        Domain name has an ASCII label ("net!") with a character not permitted.

$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net' --ns 'ns2.zonemaster.net/10.0.0.2' --ns 'ns3@.zonemaster.!net.'
Invalid name in --ns argument:
        Domain name has an ASCII label ("ns3@") with a character not permitted.
        Domain name has an ASCII label ("!net") with a character not permitted.
```

Examples of (in)valid domain names can be found in the [t/normalization.t unit test file in Engine](https://github.com/zonemaster/zonemaster-engine/blob/v4.7.3/t/normalization.t), e.g:

```
#Valid
example。com.
café.example.com

#Invalid
example。.com.
bad:%;!$.example.com.
❤️．example．com
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -18,6 +18,7 @@ use Moose;
 with 'MooseX::Getopt::GLD' => { getopt_conf => [ 'pass_through' ] };
 
 use Encode;
+use Readonly;
 use File::Slurp;
 use JSON::XS;
 use List::Util qw[max];
@@ -26,6 +27,7 @@ use Scalar::Util qw[blessed];
 use Socket qw[AF_INET AF_INET6];
 use Text::Reflow qw[reflow_string];
 use Try::Tiny;
+use Net::IP::XS;
 
 use Zonemaster::LDNS;
 use Zonemaster::Engine;
@@ -38,6 +40,9 @@ use Zonemaster::Engine::Zone;
 
 our %numeric = Zonemaster::Engine::Logger::Entry->levels;
 our $JSON    = JSON::XS->new->allow_blessed->convert_blessed->canonical;
+
+Readonly our $IPV4_RE => qr/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/;
+Readonly our $IPV6_RE => qr/^[0-9a-f:]*:[0-9a-f:]+(:[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})?$/i;
 
 STDOUT->autoflush( 1 );
 
@@ -774,8 +779,16 @@ sub add_fake_delegation {
         }
 
         if ( $ip ) {
-            $ip = new Net::IP::XS ( $ip ) or die Net::IP::XS::Error();
-            push @{ $data{ $name } }, $ip->short;
+            my $net_ip = Net::IP::XS->new( $ip );
+            if ( ( $ip =~ /($IPV4_RE)/ && Net::IP::XS::ip_is_ipv4( $ip ) )
+                or
+                 ( $ip =~ /($IPV6_RE)/ && Net::IP::XS::ip_is_ipv6( $ip ) )
+            ) {
+                push @{ $data{ $name } }, $ip;
+            }
+            else {
+                die Net::IP::XS::Error() ? "Invalid IP address in --ns argument:\n\t". Net::IP::XS::Error() ."\n" : "Invalid IP address in --ns argument.\n";
+            }
         }
         else {
             push @ns_with_no_ip, $name;
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -18,6 +18,8 @@ tests_recursive( 't' );
 # (see Zonemaster::LDNS below)
 
 requires(
+    'Readonly'           => 0,
+    'Net::IP::XS'        => 0,
     'JSON::XS'           => 0,
     'Locale::TextDomain' => 1.23,
     'MooseX::Getopt'     => 0,


lib/Zonemaster/CLI.pm
```diff
@@ -552,16 +554,18 @@ sub run {
...
-    if ( $domain =~ m/\.\./i ) {
-        die __( "The domain name contains consecutive dots.\n" );
-    }
+    ( my $errors, $domain ) = normalize_name( $domain );
 
-    $domain =~ s/\.$// unless $domain eq '.';
-    $domain = $self->to_idn( $domain );
+    if ( scalar @$errors > 0 ) {
+        my $err;
+        $err .= $_->string . "\n" for @$errors;
```
mattias-p, line 566: Maybe this line could be rewritten to avoid the use of `$_`? It can be difficult for mere mortals to spot where it was assigned.

lib/Zonemaster/CLI.pm
```diff
@@ -752,23 +756,24 @@ sub add_fake_delegation {
     foreach my $pair ( @{ $self->ns } ) {
         my ( $name, $ip ) = split( '/', $pair, 2 );
 
-        if ( not $name ) {
+        if ( $pair =~ tr/\/// > 1 or not $name ) {
```
mattias-p, line 759: This looks to me like a tightening of the validation of the IP part. Why not perform the full IP address validation here instead of just looking or extra slashes?

lib/Zonemaster/CLI.pm
```diff
@@ -752,23 +756,24 @@ sub add_fake_delegation {
...
-        if ( $name =~ m/\.\./i ) {
-            say STDERR __x( "The name of the nameserver '{nsname}' contains consecutive dots.", nsname => $name );
-            exit ( 1 );
-        }
+        ( my $errors, $name ) = normalize_name( $name );
 
-        $name =~ s/\.$// unless $name eq '.';
+        if ( scalar @$errors > 0 ) {
+            my $err = "Invalid name in --ns argument:\n" ;
+            $err .= "\t" . $_->string . "\n" for @$errors;
```
mattias-p, line 768: Maybe this line could be rewritten to avoid the use of `$_`?

mattias-p (approved):
I like this change but I have a couple of nits.

matsduf:
> $ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 '@.İabcİ.İcc'
> Domain name has an ASCII label ("@") with a character not permitted.
> Domain name has a non-ASCII label ("İabcİ") which is not a valid U-label.
> Domain name has a non-ASCII label ("İcc") which is not a valid U-label.

The problem with "İabcİ" is that it cannot be downcased in an unambiguous way. The specification says "AMBIGUOUS_DOWNCASING" (https://github.com/zonemaster/zonemaster/blob/master/docs/public/specifications/tests/RequirementsAndNormalizationOfDomainNames.md).

blacksponge:
I tested the PR with `café.test` and also got an `INVALID_ASCII` error. There seems to be missing decoding from utf8 (or whatever encoding the terminal is using), I added `Encode::decode("utf8", $domain )` before passing the domain to `nomalize_name` and it worked.

matsduf (changes requested):
Does it work correctly? I get
```
# zonemaster-cli råttgift.se
Domain name has a non-ASCII label ("råttgift") which is not a valid U-label.
```

lib/Zonemaster/CLI.pm
mattias-p: I really like the idea of passing all inputs through some kind of dedicated validation layer close to where they enter. That way it's easy to find out *if*, *where* and *how* all inputs are validated. It's also easier to provide good user feedback. E.g. instead of saying that the "ip" method can't be called on an undefined value, we could be saying that the IP address part of the `--ns` argument is invalid.

In Backend we have one validation layer for RPCAPI requests in zm-rpcapi and another one for the config file that is shared by zm-rpcapi and zm-testagent. In CLI, on the other hand, things are not so tidy and we're mixing validation and business logic. IMHO we should refactor CLI to have a proper validation layer too. In fact I'll create and issue for that.

Regarding this PR though, if you think it's out of scope for this PR to improve old error messages I'm perfectly fine with that. I think it's a great improvement that you added the validation for extra slashes.

mattias-p (approved):

blacksponge (approved):
Looks fine to me! I tested some of the test commands and it works as expected.

mattias-p (approved):
🎉

matsduf (approved):
Nice to have it in place! This gives consistent and documented behavior.

mattias-p:
I've tested this for v2023.2 and I could not find any deviations between the described behavior and the actual behavior. However, upon reflection I think the behavior could be improved in a couple of ways. I created #364 for this.

PR #359, opened 2023-11-23
More flexible --test option

mattias-p:
## Purpose

This PR implements flexible command line overriding of which test cases to run from the command line.

## Context

* This is an extension to the improvements in #333.

* Replaces #130. The point of that PR is to allow users to exclude test cases from the command line. This PR includes the same functionality using a different command line syntax.

* Leverages zonemaster/zonemaster-engine#1312 to properly handle the Basic test module (c.f. https://github.com/zonemaster/zonemaster-cli/pull/130#issuecomment-572948201).

* Handles the interaction of `--test` and `--profile` in accordance with https://github.com/zonemaster/zonemaster-cli/pull/130#issuecomment-573012219.

* For what it's worth, the idea for this design came to me in a discussion about https://github.com/zonemaster/zonemaster-engine/pull/1294.

## Changes

* The set of values accepted by `--test` is extended.

* Which test cases to include in the run is now controlled by setting the `test_cases` profile property and running Zonemaster::Engine->test_zone().

* When `--test=basic` or `--test=basic02` is specified, and basic02 considers the domain too broken to test, a CANNOT_CONTINUE message is now emitted, whereas it wasn't emitted before this PR.

## Reviewing

* Look especially for lacking test coverage.

## Testing

* Unit tests should pass.

### DIFF
--- a/lib/Zonemaster/CLI/TestCaseSet.pm
+++ b/lib/Zonemaster/CLI/TestCaseSet.pm
@@ -7,72 +7,149 @@ use Carp qw( croak );
 
 =head1 NAME
 
-    Zonemaster::CLI::TestCaseSet - A mutable set of test case names.
+    Zonemaster::CLI::TestCaseSet - Manage and modify Zonemaster test case selections
 
 =head1 SYNOPSIS
 
     use Zonemaster::CLI::TestCaseSet;
 
-    # Construct a working subset of test cases {alpha01, alpha02, alpha03,
-    # beta01} of test cases out of the full set {alpha01, alpha02, alpha03,
-    # beta01, beta02} distributed across the test modules {alpha, beta}.
-    my $working_set = Zonemaster::CLI::TestCaseSet->new(
-        \qw( alpha01 alpha02 alpha03 beta01 ),
-        {
-            alpha => \qw( alpha01 alpha02 alpha03 ),
-            beta  => \qw( beta01 beta02 ),
-        },
+    # Define the names of the available test modules and their test cases
+    my $schema = {
+        alpha => [qw( alpha01 alpha02 alpha03 )],
+        beta  => [qw( beta01 beta02 )],
+    };
+
+    # Construct an initial selection of test cases
+    my $selection = Zonemaster::CLI::TestCaseSet->new(
+        [qw( alpha01 alpha02 alpha03 beta01 )],
+        $schema,
     );
 
-    # Parse a modifier expression into a list of modifiers.
+    # Parse and apply a modifier expression
     my @modifiers = Zonemaster::CLI::TestCaseSet->parse_modifier_expr( '-alpha+alpha02' );
-
-    # Traverse the list of modifiers, chunked into (operator, term) pairs.
     while ( @modifiers ) {
-        my $op   = shift @modifiers;
-        my $term = shift @modifiers;
-
-        # Modify the working subset by applying each operator and term.
-        if ( !$working_set->apply_modifier( $op, $term ) ) {
-            die "Error: Unrecognized term '$term'.\n";
-        }
+        my ( $op, $term ) = splice @modifiers, 0, 2;
+        $selection->apply_modifier( $op, $term )
+          or die "Error: Unrecognized term '$term'.\n";
     }
 
-    # Make sure the working subset ends up in the expected state.
-    if ( join(' ', $working_set->to_list) ne 'alpha02 beta01' ) {
-        die;
-    }
+    # Output final test case selection
+    print join( ' ', $selection->to_list );    # alpha02 beta01
 
 =head1 DESCRIPTION
 
-A TestCaseSet primarily represents an immutable full set of test cases and a
-mutable subset thereof. The full set of test cases is distributed across the
-set of test modules.
+Zonemaster::CLI::TestCaseSet represents a mutable selection of test cases,
+together with an immutable schema defining available test modules and their
+associated test cases.
 
-=head2 TERM EXPANSION
+The schema is defined as a mapping of test module names to their associated test
+case names.
 
-Terms are expanded in one of three ways.
+The selection can be adjusted using modifier expressions.
+
+=head2 MODIFIER EXPRESSIONS
+
+A modifier expression describes a change to the current selection.
+Expressions combine terms using operators, e.g., C<'-alpha+alpha02'>.
+
+These operators are supported:
 
 =over 4
 
-=item The full set of test cases as provided to the TestCaseSet constructor for the current object.
+=item C<'+'> (union)
 
-Terms matching the string C<'all'>.
+Add test cases expanded from C<$term> to the current selection.
 
-=item The set of all test cases inside one test module.
+=item C<'-'> (difference)
 
-Terms matching the name of a test module.
+Remove test cases expanded C<$term> from the current selection.
 
-=item The singleton set of a single test cases
+=item C<''> (replace)
 
-Terms matching the name of a test case (e.g. "Case10") or the concatenation of a test module,
-a slash and a test cases belonging to that test module (e.g. "Case/Case10").
+Replace the entire current selection with the test cases expanded from C<$term>.
 
 =back
 
-Term names are matched case insensitively.
+Terms expand into sets of test cases in one of three ways:
 
-=head1 SUBROUTINES
+=over 4
+
+=item C<all>
+
+Expands to all available test cases defined by the schema.
+
+=item Test module name
+
+Expands to all test cases associated with the test module.
+
+=item Test case name
+
+Expands directly to the specified test case itself.
+Test cases may be specified plainly (e.g., C<Case10>) or fully qualified
+(module/testcase, e.g., C<Case/Case10>).
+
+=back
+
+Term matching is case-insensitive.
+
+=cut
+
+=head1 CONSTRUCTORS
+
+=head2 new( $selection, $schema )
+
+Construct a new TestCaseSet object.
+
+=over 4
+
+=item C<$selection> (arrayref)
+
+Initial selection of test case names.
+
+=item C<$schema> (hashref)
+
+A hash mapping test module names to arrays of their associated test case names.
+
+=back
+
+Dies if:
+- Any test case name in C<$schema> is repeated.
+- C<$selection> contains names not found in C<$schema>.
+
+=cut
+
+sub new {
+    my ( $class, $selection, $schema ) = @_;
+
+    my %cases = map { lc $_ => 1 } map { @{$_} } values %$schema;
+    for my $case ( @$selection ) {
+        if ( !exists $cases{ lc $case } ) {
+            croak "Unrecognized initial test case '$case'";
+        }
+    }
+
+    my $obj = {
+        _selection => { map { lc $_ => 1 } @$selection },
+        _terms     => _get_schema_terms( $schema ),
+    };
+
+    bless $obj, $class;
+
+    return $obj;
+}
+
+=head1 CLASS METHODS
+
+parse_modifier_expr( $modifier_expr )
+
+Parse a string containing a modifier expression and returns a list of
+alternating operators and terms.
+
+The returned list always starts with an operator.
+
+For example, parsing C<'-alpha+beta02'> returns:
+
+    ('-', 'alpha', '+', 'beta02')
 
 =cut
 
@@ -90,94 +167,46 @@ sub parse_modifier_expr {
     return @modifiers;
 }
 
-=head1 CONSTRUCTORS
-
-=head2 new()
-
-In the full set of test cases, methods names must not share the same name as
-other test cases or test modules.
-
-=cut
-
-sub new {
-    my ( $class, $initial_methods, %all_methods ) = @_;
-
-    my %flattened_methods = map { $_ => 1 } map { @{$_} } values %all_methods;
-    for my $method ( @$initial_methods ) {
-        if ( !exists $flattened_methods{$method} ) {
-            croak "Unrecognized initial method '$method'";
-        }
-    }
-
-    my $obj = {
-        _cur_methods      => { map { $_ => 1 } @$initial_methods },
-        _all_term_methods => _get_all_term_methods( \%all_methods ),
-    };
-
-    bless $obj, $class;
-
-    return $obj;
-}
-
 =head1 INSTANCE METHODS
 
-=head2 apply_modifier()
+=head2 apply_modifier( $operator, $term )
 
-Update the working subset.
+Update the selection using the given operator and term.
 
-The given operator is applied to two operands and the result is assigned to the
-working subset. The left hand side operand is the current value of the working
-subset. The right hand side operand is calculated by L<expanding|/"TERM
-EXPANSION"> the given term to a subset of test cases.
+Returns true if successful, or false if the term could not be expanded based on
+the schema.
 
-Three operators are supported.
+Dies if the operator is invalid.
 
-=over 4
+=head3 Example:
 
-=item C<'+'>
-
-Returns the union of the left and right hand side operands
-
-=item C<'-'>
-
-Returns the set difference of the left and right hand side operands
-
-=item C<''>
-
-Ignores the left hand side operand and returns the right hand side operand.
-
-=back
-
-Returns true if the operation is successful.
-
-Returns false if the term could not be expanded.
-
-Dies if the operator is not recognized.
+    $selection->apply_modifier('+', 'beta') 
+        or die "Unrecognized term";
 
 =cut
 
 sub apply_modifier {
     my ( $self, $op, $term ) = @_;
 
-    my $methods_ref = $self->{_all_term_methods}{ lc $term };
+    my $cases_ref = $self->{_terms}{ lc $term };
 
-    if ( !defined $methods_ref ) {
+    if ( !defined $cases_ref ) {
         return 0;
     }
 
     if ( $op eq '' ) {
-        $self->{_cur_methods} = {};
+        $self->{_selection} = {};
         $op = '+';
     }
 
     if ( $op eq '-' ) {
-        for my $method ( @$methods_ref ) {
-            delete $self->{_cur_methods}{$method};
+        for my $case ( @$cases_ref ) {
+            delete $self->{_selection}{$case};
         }
     }
     elsif ( $op eq '+' ) {
-        for my $method ( @$methods_ref ) {
-            $self->{_cur_methods}{$method} = 1;
+        for my $case ( @$cases_ref ) {
+            $self->{_selection}{$case} = 1;
         }
     }
     else {
@@ -187,47 +216,53 @@ sub apply_modifier {
     return 1;
 } ## end sub apply_modifier
 
+=head2 to_list
+
+Return a lowercase list of the currently selected test case names.
+
+=cut
+
 sub to_list {
     my ( $self ) = @_;
 
-    return sort keys %{ $self->{_cur_methods} };
+    return sort keys %{ $self->{_selection} };
 }
 
-sub _get_all_term_methods {
-    my ( $all_methods ) = @_;
+sub _get_schema_terms {
+    my ( $schema ) = @_;
 
     my $terms = {};
     $terms->{all} = [];
 
-    for my $module ( keys %$all_methods ) {
+    for my $module ( keys %$schema ) {
         if ( lc $module eq 'all' ) {
-            croak "module name must not be 'all'";
+            croak "test module name must not be 'all'";
         }
         if ( $module =~ qr{/} ) {
-            croak "module name contains forbidden character '/': '$module'";
+            croak "test module name contains forbidden character '/': '$module'";
         }
         if ( exists $terms->{ lc $module } ) {
-            croak "found module with same name as another method or module: '$module'";
+            croak "found test module with same name as another test case or test module: '$module'";
         }
         $terms->{ lc $module } = [];
-        for my $method ( @{ $all_methods->{$module} } ) {
-            if ( lc $method eq 'all' ) {
-                croak "method name must not be 'all'";
+        for my $case ( @{ $schema->{$module} } ) {
+            if ( lc $case eq 'all' ) {
+                croak "test case name must not be 'all'";
             }
-            if ( $method =~ qr{/} ) {
-                croak "method name contains forbidden character '/': '$method'";
+            if ( $case =~ qr{/} ) {
+                croak "test case name contains forbidden character '/': '$case'";
             }
-            if ( exists $terms->{ lc $method } ) {
-                croak "found method with same name as another method or module: '$method'";
+            if ( exists $terms->{ lc $case } ) {
+                croak "found test case with same name as another test case or test module: '$case'";
             }
-            $terms->{ lc $method } = [$method];
-            $terms->{ lc "$module/$method" } = [$method];
-            push @{ $terms->{ lc $module } }, $method;
-            push @{ $terms->{all} },          $method;
+            $terms->{ lc $case } = [$case];
+            $terms->{ lc "$module/$case" } = [$case];
+            push @{ $terms->{ lc $module } }, $case;
+            push @{ $terms->{all} },          $case;
         }
-    } ## end for my $module ( keys %$all_methods)
+    } ## end for my $module ( keys %$schema)
 
     return $terms;
-} ## end sub _get_all_term_methods
+} ## end sub _get_schema_terms
 
 1;
--- a/lib/Zonemaster/CLI/TestCaseSet.pm
+++ b/lib/Zonemaster/CLI/TestCaseSet.pm
@@ -1,0 +1,233 @@
+package Zonemaster::CLI::TestCaseSet;
+use 5.014;
+use warnings;
+use utf8;
+
+use Carp qw( croak );
+
+=head1 NAME
+
+    Zonemaster::CLI::TestCaseSet - A mutable set of test methods names.
+
+=head1 SYNOPSIS
+
+    use Zonemaster::CLI::TestCaseSet;
+
+    # Construct a working subset of test methods {alpha01, alpha02, alpha03,
+    # beta01} of test methods out of the full set {alpha01, alpha02, alpha03,
+    # beta01, beta02} distributed across the test modules {alpha, beta}.
+    my $working_set = Zonemaster::CLI::TestCaseSet->new(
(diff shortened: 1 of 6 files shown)


matsduf:
>     * Before this PR CLI would call either `test_zone`, `test_module` or `test_method` in Zonemaster::Engine to run the test. After this PR CLI always calls `test_zone` and uses the `test_cases` profile property to control which tests are run. This means that the Basic tests are always included even when a single test case is specified.
> 
>     * Since the semantics of specifying a single test case or single test module is changed so that the Basic tests are included, an argument could be made that this is a breaking change, but I believe an argument could also be made that it's still just an addition.

That is a bad change. When working with test zones you want to be able to run just the specific test case and nothing else. 

Secondly, with that change, if Basic fails completely no other test case is run, and that is not what we want when testing and creating test zones.

I am against a change that makes it impossible to run just the selected test case.

mattias-p:
All I'm hearing is "no" and "bad". Is there anything about this proposal that you do like?

Also, I'm trying to understand your objection here. AFAICT you're saying that when a user requests that a single test case is run on a domain that doesn't fulfill the requirements posed by Basic, the requested test case still needs to run. Is that correct? And if so, could you explain why this is important, because it's not obvious to me.

matsduf:
> All I'm hearing is "no" and "bad". Is there anything about this proposal that you do like?

It would be good to be able to exclude test cases without creating a new profile. It would be good to be able to select two or more test cases and they are run without overlap.

Sometimes it is probably good that Basic is run before a selected test.

> Also, I'm trying to understand your objection here. AFAICT you're saying that when a user requests that a single test case is run on a domain that doesn't fulfill the requirements posed by Basic, the requested test case still needs to run. Is that correct? And if so, could you explain why this is important, because it's not obvious to me.

I want to see what happens in that test case for the test zone even when there are errors. And I want minimal output without the extra output from Basic when testing test zones. Sometimes I have to run with `--level debug` which gives many lines, which will be anymore if Basic is run.

Give the possibility to run a test without adding Basic.

If Basic is not in the profile, why should Basic still be run? The magic could rather be to put the Basic test cases in the profile first, if there are any.

mattias-p:
> I want to see what happens in that test case for the test zone even when there are errors. And I want minimal output without the extra output from Basic when testing test zones. Sometimes I have to run with --level debug which gives many lines, which will be anymore if Basic is run.

> If Basic is not in the profile, why should Basic still be run? The magic could rather be to put the Basic test cases in the profile first, if there are any.

Ok. Yeah, I see how this could be very valuable in development contexts. And I also don't particularly like that Basic is forced.

I have an idea of how we could simplify things in Engine in a way that supports this. I think I'll make a draft PR there too and we can let this PR rest until Engine is in a better shape to support it.

matsduf:
@mattias-p, is this still relevant or can it be closed now?

mattias-p:
This depends on #371. I plan to rebase this PR the documentation improvements after #371 PR have been merged. It seems somewhat unreasonable to get this merged for 2024.2 so I postponed it.

mattias-p:
I rebased this onto latest develop and fixed some typos added some unit tests. I also tightened some validations for sanity reasons, but that shouldn't have any effect on zonemaster-cli behavior.

matsduf:
This is not true anymore. Please update description.

>     * This PR contains the commits from [Organize --help text #389](https://github.com/zonemaster/zonemaster-cli/pull/389), which should be merged before this one.

matsduf:
From `man zonemaster-cli`:

```
   Testing Options
       --test=TESTCASE, --test=TESTMODULE
           Limit the testing suite to run only the specified tests.  (default:
           the empty string)

           EXPR is an expression specifying or modifying the set of tests to
           run.  Here are some examples:
```
Shouldn't it be `--test=EXPR`?

matsduf:
In the man page I find the following two examples,

```
           --test=all-nameserver+nameserver03
               runs all test cases of all test modules except for the
               Nameserver test module in which only the Nameserver03 test case
               is run.

           --test=-dnssec
               removes all test cases of the DNSSEC module from the set of
               tests that would otherwise have run.
```

Does `--test=-dnssec` and `--test=all-dnssec` mean the same thing?

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+=begin :man
+
+EXPR is an expression specifying or modifying the set of tests to run.
+Here are some examples:
+
+=over 4
+
+=item --test=B<nameserver>
+
+runs the entire Nameserver test module, and nothing else.
```
matsduf, line 128:
```suggestion
runs the entire Nameserver test module, including any test cases
disabled by the profile, but nothing else.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
-test case.
-This option is case-insensitive.
+(default: the empty string)
+
+=begin :man
+
+EXPR is an expression specifying or modifying the set of tests to run.
+Here are some examples:
+
+=over 4
```
matsduf, line 124:
```suggestion
=over 4

=item --test=all

runs all defined test cases including any test cases 
disabled by the profile.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+runs the entire Nameserver test module, and nothing else.
+
+=item --test=B<nameserver01>
+
+runs the Nameserver01 test case, and nothing else.
+
+=item --test=B<dnssec-dnssec05>
+
+runs the entire DNSSEC test module except the DNSSEC05 test case, and nothing
+else.
```
matsduf, line 137:
```suggestion
runs the entire DNSSEC test module including any test cases
disabled by the profile, except the DNSSEC05 test case,
and nothing else.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+otherwise have run.
+
+=item --test=B<+syntax01+syntax02>
+
+adds the Syntax01 and Syntax02 test cases to the set of tests that would
+otherwise have run.
+
+=item --test=
+
+(i.e., the empty string) leaves the set of tests to run unmodified.
```
matsduf, line 156:
```suggestion
(i.e., the empty string) leaves the set of tests to run unmodified,
which means that all test cases enabled by the profile will
be run.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
 =item B<--test>=TESTCASE, B<--test>=TESTMODULE
```
matsduf, line 114:
```suggestion
=item B<--test>=TESTCASE, B<--test>=TESTMODULE, B<--test>=EXPR
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+
+=item --test=B<all-nameserver+nameserver03>
+
+runs all test cases of all test modules except for the Nameserver test module in
+which only the Nameserver03 test case is run.
+
+=item --test=B<-dnssec>
+
+removes all test cases of the DNSSEC module from the set of tests that would
+otherwise have run.
```
matsduf, line 147:
```suggestion
otherwise have run, i.e. the test cases enabled by the profile.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+
+=item --test=B<+syntax01+syntax02>
+
+adds the Syntax01 and Syntax02 test cases to the set of tests that would
+otherwise have run.
```
matsduf, line 152:
```suggestion
otherwise have run, i.e. the test cases enabled by the profile.
```

script/zonemaster-cli
```diff
@@ -114,14 +114,50 @@ Print the effective profile used in JSON format, then exit.
...
+
+=item --test=B<dnssec-dnssec05>
+
+runs the entire DNSSEC test module except the DNSSEC05 test case, and nothing
+else.
+
+=item --test=B<all-nameserver+nameserver03>
+
+runs all test cases of all test modules except for the Nameserver test module in
+which only the Nameserver03 test case is run.
```
matsduf, line 142:
```suggestion
runs all defined test cases of all defined test modules (disregardning the profile)
except for the Nameserver test module in which only the Nameserver03 test
case is run.
```

matsduf (changes requested):

matsduf:
Nice to have it consistent. Nice to be able to control what test cases to run without creating a custom profile.

mattias-p:
> Shouldn't it be --test=EXPR?

Fixing.

> Do `--test=-dnssec` and `--test=all-dnssec` mean the same thing?
> 
> ---> No
> 
> Do `--test=` and `--test=all` mean the same thing?
> 
> ---> No

Agreed. In case you modify the `test_cases` property using --profile they won't be the same.

mattias-p:
I think the current implementation is flawed regarding the interpretation of this command sequence:
```sh
printf "%s\n" "--test=-delegation01" > ~/.zonemaster/cli.args
zonemaster-cli example.com
zonemaster-cli --test=-delegation02 example.com
```

With the current proposal, in the first command delegation01 would be disabled, but in the second command it would be enabled again. This is because the second command would effectively be `zonemaster-cli --test=-delegation01 --test=-delegation02 example.com` and the second --test would override the first one. It would make much more sense if both --test arguments were applied as they appear on the command line. `zonemaster-cli --test=-delegation01 --test=basic01` would still mean execute basic01 and nothing else.

script/zonemaster-cli
mattias-p: I agree that the documentation should make this clear, but I don't think it needs to be repeated for every example. I'll add a separate paragraph clarifying how it interacts with other options.

matsduf:
> I think the current implementation is flawed regarding the interpretation of this command sequence:
> 
> ```shell
> printf "%s\n" "--test=-delegation01" > ~/.zonemaster/cli.args
> zonemaster-cli example.com
> zonemaster-cli --test=-delegation02 example.com
> ```
> 
> With the current proposal, in the first command delegation01 would be disabled, but in the second command it would be enabled again. This is because the second command would effectively be `zonemaster-cli --test=-delegation01 --test=-delegation02 example.com` and the second --test would override the first one. It would make much more sense if both --test arguments were applied as they appear on the command line. `zonemaster-cli --test=-delegation01 --test=basic01` would still mean execute basic01 and nothing else.

I think it is correct that `--test` on the command line overrides `--test` in the configuration file. That is how it works in other cases.

mattias-p:
We have two kinds of option today: overwriting ones and aggregate ones. Most of our options overwrite, but --test, --ns and --ds aggregate. I'm arguing that it makes more sense to keep --test as an aggregating option to change it into an overwriting one. I.e., `--test=basic01 --test=+basic02` should be equivalent to `--test=basic01+basic02`, not to `--test=+basic02`.

mattias-p:
I pushed an update makes --test aggregate again. It's in a separate commit in case you want to play with both the overwriting version and the aggregating one. The man page needs an update to clarify the semantics in of repeated --test options but I haven't got around to that.

I haven't forgotten your unresolved review comments, @matsduf. It's just that before I take care of them I want to settle this overwriting/aggregating business.

matsduf:
> We have two kinds of option today: overwriting ones and aggregating ones. Most of our options overwrite, but --test, --ns and --ds aggregate. I'm arguing that it makes more sense to keep --test as an aggregating option than to change it into an overwriting one. I.e., `--test=basic01 --test=+basic02` should be equivalent to `--test=basic01+basic02`, not to `--test=+basic02`.

There is a clear difference between `--ns` and `--ds`, on one hand, and  `--test`, on the other hand.

Options `--ns` and `--ds` are typically used for a specific zone. It is unlikely that you would put that into the the configuration file `cli.args`. Therefor it works to have them aggregating. On the other hand, it would also work to change them so that only the last one is used, but so that you can specify multiple NS or DS, respectively, for the option, e.g. with a `+` sign or in quotemarks.

Option `--test` can be used in `cli.args` instead of creating a custom profile. Then it should be possible to override that with the settings on the command line.

mattias-p:
I've updated the documentation, applied perltidy and rebased onto current develop. Please review again.

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,233 @@
...
+mutable subset thereof. The full set of test methods is distributed across the
+set of test modules.
+
+=head2 TERM EXPANSION
+
+Terms are expanded in one of three ways.
+
+=over 4
+
+=item The full set of all test methods.
```
matsduf: All test methods in the default profile?

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,233 @@
+package Zonemaster::CLI::TestCaseSet;
+use 5.014;
+use warnings;
+use utf8;
+
+use Carp qw( croak );
+
+=head1 NAME
+
+    Zonemaster::CLI::TestCaseSet - A mutable set of test methods names.
```
matsduf: It is confusing when it says "test methods" when else in Zonemaster refer to "test cases".

```suggestion
    Zonemaster::CLI::TestCaseSet - A mutable set of test method names (test case names).
```

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,233 @@
...
+    # Make sure the working subset ends up in the expected state.
+    if ( join(' ', $working_set->to_list) ne 'alpha02 beta01' ) {
+        die;
+    }
+
+=head1 DESCRIPTION
+
+A TestCaseSet primarily represents an immutable full set of test methods and a
+mutable subset thereof. The full set of test methods is distributed across the
+set of test modules.
```
matsduf:
```suggestion
A TestCaseSet primarily represents an immutable full set of test methods (test cases) and a
mutable subset thereof. The full set of test methods (test cases) is distributed across the
set of test modules.
```

script/zonemaster-cli
```diff
@@ -111,17 +108,65 @@ Print the effective profile used in JSON format, then exit.
...
+EXPR is composed of terms (individual test cases, test modules, or C<all>) and
+operators (C<+> to add, C<-> to remove).
+For details, see C<man zonemaster-cli>.
+
+=begin :man
+
+An EXPR starting with a term (e.g., C<all>, C<nameserver01>) completely replaces
+the current value of C<test_cases>.
+An EXPR starting with C<+> or C<-> modifies the existing C<test_cases>.
+An empty EXPR (C<--test=>) leaves C<test_cases> unchanged.
```
matsduf: Can you clearify what it means to "leave 'test_cases' unchanged'?

script/zonemaster-cli
```diff
@@ -111,17 +108,65 @@ Print the effective profile used in JSON format, then exit.
...
+
+removes all DNSSEC test cases.
+
+=item --test=B<+syntax01+syntax02>
+
+adds Syntax01 and Syntax02 test cases.
+
+=item --test=
+
+leaves the existing C<test_cases> unmodified.
```
matsduf: See above.

script/zonemaster-cli
```diff
@@ -111,17 +108,65 @@ Print the effective profile used in JSON format, then exit.
...
+
+An EXPR starting with a term (e.g., C<all>, C<nameserver01>) completely replaces
+the current value of C<test_cases>.
+An EXPR starting with C<+> or C<-> modifies the existing C<test_cases>.
+An empty EXPR (C<--test=>) leaves C<test_cases> unchanged.
+
+Multiple C<--test> options may be provided; they are applied sequentially.
+When used with C<--profile>, that option is applied first.
+
+Use C<--list-tests> to list all available test cases and test modules.
```
matsduf:
```suggestion
Use C<--list-tests> to list all available test cases and test modules in the default profile.
```

script/zonemaster-cli
```diff
@@ -111,17 +108,65 @@ Print the effective profile used in JSON format, then exit.
...
+
+Examples:
+
+=over 4
+
+=item --test=B<all>
+
+selects all available test cases.
```
matsduf:
```suggestion
selects all available test cases in the default profile.
```

matsduf (changes requested):
I have tested and it seems to work as I think it was meant to work.

lib/Zonemaster/CLI/TestCaseSet.pm
mattias-p: Good idea. Though I'll express it in another way.

lib/Zonemaster/CLI/TestCaseSet.pm
mattias-p: I agree. I'll just avoid referring to test methods altogether.

lib/Zonemaster/CLI/TestCaseSet.pm
mattias-p: On second thought, no. The TestCaseSet doesn't have any concept of profiles at all. It just works with whatever you tell it in the constructor.

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,233 @@
...
+mutable subset thereof. The full set of test cases is distributed across the
+set of test modules.
+
+=head2 TERM EXPANSION
+
+Terms are expanded in one of three ways.
+
+=over 4
+
+=item The full set of test cases.
```
matsduf:
```suggestion
=item The full set of test cases in the default profile.
```

matsduf (changes requested):

script/zonemaster-cli
mattias-p: Fixed.

script/zonemaster-cli
mattias-p: I changed how I refer to test_cases.

lib/Zonemaster/CLI/TestCaseSet.pm
mattias-p: As I replied before, the TestCaseSet doesn't have any concept of profiles at all. It just works with whatever you tell it in the constructor.

matsduf: When you select "all" you get all test cases in the default profile.

Could you give a better explanation and description of `TestCaseSet.pm`? I do not understand.

script/zonemaster-cli
```diff
@@ -111,17 +108,66 @@ Print the effective profile used in JSON format, then exit.
...
+operators (C<+> to add, C<-> to remove).
+For details, see C<man zonemaster-cli>.
+
+=begin :man
+
+An EXPR starting with a term (e.g., C<all>, C<nameserver01>) completely replaces
+the selection (i.e., the value of the C<test_cases> profile property).
+An EXPR starting with an operator applies a delta to the selection.
+An empty EXPR (C<--test=>) leaves the selection unchanged (formally, it is
+updated by applying the identity transformation).
```
matsduf: What does "leaves the selection unchanged" mean? For what can I use that?

script/zonemaster-cli
```diff
@@ -111,17 +108,66 @@ Print the effective profile used in JSON format, then exit.
...
+
+removes all DNSSEC test cases from the selection.
+
+=item --test=B<+syntax01+syntax02>
+
+adds the Syntax01 and Syntax02 test cases to the selection.
+
+=item --test=
+
+leaves the selection unchanged.
```
matsduf: When do I want to use that feature? What does it give me?

matsduf (changes requested):

lib/Zonemaster/CLI/TestCaseSet.pm
mattias-p: > When you select "all" you get all test cases in the default profile.

This true if and only if you provide "all test cases in the default profile" to the TestCaseSet constructor. In other words, it depends on how the module is used. The documentation of a module should not depend on how that module is being used.

script/zonemaster-cli
mattias-p: `--test=` is useful the same way `x + 0` is useful. I don't expect you to type it out on the command line, but it could make life easier if you want to call zonemaster-cli programmatically.

matsduf: Can you please clarify that, preferably below when it listed on its own line.

script/zonemaster-cli
matsduf: Can you please clarify here that it will not do anything?

lib/Zonemaster/CLI/TestCaseSet.pm
matsduf: Can you please give a more explicit description of the module? I still do not understand it.

matsduf:
```suggestion
=item The full set of test cases as provided to the TestCaseSet constructor for the current object.
```

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,233 @@
...
+Terms matching the string C<'all'>.
+
+=item The set of all test cases inside one test module.
+
+Terms matching the name of a test module.
+
+=item The singleton set of a single test cases
+
+Terms matching the name of a test case or the concatenation of a test module,
+a slash and a test cases belonging to that test module.
```
matsduf:
```suggestion
Terms matching the name of a test case (e.g. "Case10") or the concatenation of a test module,
a slash and a test cases belonging to that test module (e.g. "Case/Case10").
```

script/zonemaster-cli
matsduf:
```suggestion
=item --test=

empty value is valid, but does not change the selection of test cases
```

mattias-p:
> Can you please give a more explicit description of the module? I still do not understand it.

I read through the module documentation again and I agree there was some room for improvement. So I worked through it again.

mattias-p:
I've rebased onto current develop and also improved the documentation.

Please review again.

script/zonemaster-cli
```diff
@@ -111,17 +108,67 @@ Print the effective profile used in JSON format, then exit.
...
+
+=begin :man
+
+An EXPR starting with a term (e.g., C<all>, C<nameserver01>) I<replaces>
+the current selection.
+An EXPR starting with an operator (C<+> or C<->) I<modifies> the current selection.
+An empty EXPR (C<--test=>) keeps the current selection unchanged (i.e., it is a no-op).
+
+Multiple C<--test> options may be specified; they are applied in the order they are given.
+When used together with the C<--profile>, that option is applied first.
```
matsduf: To make it clearly unambiguous:

```suggestion
When used together with the C<--profile>, the profile option is applied first.
```

script/zonemaster-cli
```diff
@@ -111,17 +108,67 @@ Print the effective profile used in JSON format, then exit.
...
+The EXPR is a flexible expression consisting of terms (individual test cases, test
+modules, or all) and operators (C<+> to add, C<-> to remove).
+For details, see C<man zonemaster-cli>.
+
+=begin :man
+
+An EXPR starting with a term (e.g., C<all>, C<nameserver01>) I<replaces>
+the current selection.
+An EXPR starting with an operator (C<+> or C<->) I<modifies> the current selection.
+An empty EXPR (C<--test=>) keeps the current selection unchanged (i.e., it is a no-op).
```
matsduf:
```suggestion
An empty EXPR (C<--test=>) does not modify the selection (i.e., it is a no-op).
```

script/zonemaster-cli
```diff
@@ -111,17 +108,67 @@ Print the effective profile used in JSON format, then exit.
...
+
+Use C<--list-tests> to list all test cases and test modules in the default profile.
+
+Examples:
+
+=over 4
+
+=item --test=B<all>
+
+selects all test cases in the default profile.
```
matsduf:
```suggestion
selects all test cases in the default profile, i.e. overrides any custom profile.
```

script/zonemaster-cli
```diff
@@ -111,17 +108,67 @@ Print the effective profile used in JSON format, then exit.
...
+
+=item --test=B<nameserver>
+
+selects all test cases in the Nameserver test module.
```
matsduf:
```suggestion
selects all test cases in the Nameserver test module in the default profile.
```

matsduf (commented):
This fine, but I have some comments on the documentation to make is clearer. The functions work well.

mattias-p:
@matsduf, I applied your suggestions. Please review again.

matsduf (dismissed):

lib/Zonemaster/CLI/TestCaseSet.pm
```diff
@@ -0,0 +1,271 @@
...
+            delete $self->{_selection}{$case};
+        }
+    }
+    elsif ( $op eq '+' ) {
+        for my $case ( @$cases_ref ) {
+            $self->{_selection}{$case} = 1;
+        }
+    }
+    else {
+        croak "Unrecognized operator '$op'";
```
marc-vanderwal, line 216: This line is the only thing not covered by the existing unit tests.

t/test_case_set.t
```diff
@@ -0,0 +1,303 @@
...
+                    eq_or_diff $actual, $case->{expect_ok}, "result";
+                }
+                else {
+                    eq_or_diff $actual, undef, "no result";
+                }
+            };    ## end sub
+        } ## end for my $case ( @cases )
+    };    ## end 'new' => sub
+
+    subtest 'apply_modifier' => sub {
```
marc-vanderwal, line 214: There should be a test case checking whether there is a `croak` if passing an illegal modifier.

marc-vanderwal (commented):
Looks fine.

I’ve run the test suite through Devel::Cover and found a single line in `TestCaseSet.pm` that wasn’t covered by the test suite. Otherwise it’s 100% covered.

mattias-p:
I've rebased and added a unit test for invalid operators. The unit test is slightly messy, but I didn't know how to do it more cleanly. Please review again.

marc-vanderwal (approved):

matsduf (approved):

marc-vanderwal:
Successfully release tested on Rocky Linux 9.

Unit tests pass, and I did a few test runs with varying `--test` command line arguments, using the new expression syntax this PR introduces. It seems to work as intended.

PR #371, opened 2024-05-28
Make usage documentation consistent and also demoose

### LINKED ISSUE #68: Inconsistent CLI options documentation
The command line options are documented in two different ways. The two different documentations can be show with the commands `zonemaster-cli --help` and `perldoc zonemaster-cli` respectively. Both commands should output the same documentation.

mattias-p:
## Purpose

This PR kills two birds with one stone:
 * It fixes the discrepancy between perldoc zonemaster-cli and zonemaster-cli --help, and
 * it migrates away from Moose and Moosex::Getopt::GLD.

## Context

* Fixes #68.
* I've tried to keep this PR conservative with regards to improving the actual help text. I have a lot of improvements in mind but I'm deferring them to one or more followup PRs. See #389.

## Changes

Application:
 * Instead of using Moose and Moosex::Getopt::GLD we now Getopt::Long and Pod::Usage which are both Perl core modules.
 * I merged the best parts of the --help and perldoc usage texts. As I worked through them I realized there was a lot of room for further improvement. I tried to be conservative making improvements but in some cases I couldn't help myself.

Tests:
 * A unit test is added to make sure --help work.
 * A unit test is added to make sure all POD is syntactically correct.

Documentation:
 * The --help text is longer now and formatted differently. This because we're using a different method of generating it and the details are taken from POD instead of Moose attributes.

## Reviewing this PR

* Inspect the help text and man page of both this branch and the develop branch looking details that have gone missing.

## Testing this PR

* Unit tests should pass.
* Perform a limited amount of exploratory testing for edge cases like missing required arguments, unknown options, invalid arguments, or conflicting flags.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -1,7 +1,8 @@
 # Brief help module to define the exception we use for early exits.
 package Zonemaster::Engine::Exception::NormalExit;
-use Moose;
-extends 'Zonemaster::Engine::Exception';
+use 5.014002;
+use warnings;
+use parent 'Zonemaster::Engine::Exception';
 
 # The actual interesting module.
 package Zonemaster::CLI;
@@ -14,14 +15,14 @@ use warnings;
 use version; our $VERSION = version->declare( "v7.0.0" );
 
 use Locale::TextDomain 'Zonemaster-CLI';
-use Moose;
-with 'MooseX::Getopt::GLD' => { getopt_conf => [ 'pass_through' ] };
 
 use Encode;
 use File::Slurp;
+use Getopt::Long qw[GetOptionsFromArray :config bundling no_auto_abbrev];
 use JSON::XS;
 use List::Util qw[max uniq];
 use Net::IP::XS;
+use Pod::Usage;
 use POSIX qw[setlocale LC_MESSAGES LC_CTYPE];
 use Readonly;
 use Scalar::Util qw[blessed];
@@ -37,6 +38,7 @@ use Zonemaster::Engine::Validation qw[validate_ipv4 validate_ipv6];
 
 our %numeric = Zonemaster::Engine::Logger::Entry->levels;
 our $JSON    = JSON::XS->new->allow_blessed->convert_blessed->canonical;
+our $SCRIPT  = $0;
 
 Readonly our $EXIT_SUCCESS       => 0;
 Readonly our $EXIT_GENERIC_ERROR => 1;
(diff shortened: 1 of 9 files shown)


lib/Zonemaster/CLI.pm
```diff
@@ -1,7 +1,8 @@
 # Brief help module to define the exception we use for early exits.
 package Zonemaster::Engine::Exception::NormalExit;
-use Moose;
-extends 'Zonemaster::Engine::Exception';
+use 5.014002;
+use warnings;
+our @ISA = qw( Zonemaster::Engine::Exception );
```
marc-vanderwal, line 5: Why not `use parent Zonemaster::Engine::Exception`?

mattias-p: Fixing.

matsduf:
I guess `.travis.yml` should also be updated to match the update. The installation instruction also needs an update.

lib/Zonemaster/CLI.pm
```diff
@@ -389,6 +381,22 @@ sub run {
         Zonemaster::Engine::Profile->effective->merge( $profile );
     }
 
+    if ( defined $self->sourceaddr4 ) {
+        eval { Zonemaster::Engine::Profile->effective->set( q{resolver.source4}, $self->sourceaddr4 ); };
+        if ( $@ ) {
+            say STDERR __x( "Error: invalid value for --sourceaddr4: {reason}", reason => $@ );
```
marc-vanderwal, line 228: Is the `reason` argument also translated according to the user’s locale?

matsduf: It depends on the OS and the OS locale setting if `$@` is translated or not, doen't it?

mattias-p: No, it's a plain English string. I created zonemaster/zonemaster-engine#1375 to internationalize it.

mattias-p: @matsduf But I believe you're right that errors reported by the OS (and possibly other dependencies) may be internationalized.

mattias-p:
I've split out #381 from this PR and rebased this one on top of that one.

lib/Zonemaster/CLI.pm
```diff
@@ -626,39 +458,39 @@ sub run {
...
         } or do {
             print STDERR $@;
             return $EXIT_USAGE_ERROR;
         };
     }
 
-    if ( $self->ds and @{ $self->ds } ) {
-        $self->check_fake_ds( $domain );
+    if ( @opt_ds ) {
+        check_fake_ds( $domain, @opt_ds );
```
mattias-p, line 493: I immediately found a bug here that wasn't caught by the unit tests. I'll see if I can add a unit test to cover it or otherwise I'll add an item to the testing section of the PR description.

matsduf:
Have you looked at #388?

matsduf:
The man page starts with
```
zonemaster-cli --test=delegation/delegation01 --level=debug zonemaster.net
```
The easier and preferred format is
```
zonemaster-cli --test=delegation01 --level=debug zonemaster.net
```

matsduf:
`zonemaster-cli -h` gives brief information now, and that is fine, but I do not get any information on how to get full information. `zonemaster-cli --help` gives brief information. Is it possible to give it paged like a man page? `perldoc Zonemaster::CLI` gives no information. Maybe refer to `man zonemaster-cli` and `zonemaster-cli --help`?

matsduf:
I think `zonemaster-cli` with no argument should refer to some documentation like "--help".

mattias-p:
> The man page starts with
> 
> ```
> zonemaster-cli --test=delegation/delegation01 --level=debug zonemaster.net
> ```
> 
> The easier and preferred format is
> 
> ```
> zonemaster-cli --test=delegation01 --level=debug zonemaster.net
> ```

As I state in the PR description I'm deferring all improvements to the text. If you look at #389 you'll see that I update the SYNOPSIS section so that it declares the command line syntax instead of just a selection of examples.

matsduf (approved):

matsduf:
@mattias-p, please look at issue #388. I think you are behind that code.

marc-vanderwal (approved):
Looks good to me.

marc-vanderwal:
Release testing for 2024.2: I have some remarks, but there are no show-stoppers.

What works:

 * `zonemaster-cli --json-stream --no-json` is rejected, as it should;
 * `zonemaster-cli --json-translate` without `--json` nor `--json-stream` gives a deprecation warning and a warning that `--json-stream` has no effect;
 * invalid values for `--level` and `--stop-level` are rejected;
 * erroneous values for `--ns` or `--ds` (that do not respect the prescribed syntax) are rejected.

But there are some things worth looking at:

 * A command line such as `zonemaster-cli --help --list-tests --version --dump-profile` doesn’t make much sense; maybe these four options should be all made mutually exclusive to each other?
 * In the synopsis, one line says `zonemaster-cli [OPTIONS] --dump-profile`. Except from `--profile=FILE`, are there any other options that could make sense when combined with `--dump-profile`? For example, `zonemaster-cli --dump-profile --no-json` still gives me JSON output because `--no-json` is not applicable to the output caused by `--dump-profile`.
 * The error messages displayed on erroneous values for `--level` and `--stop-level` are inconsistent:
```
$ zonemaster-cli --stop-level=INVALID
Failed to recognize stop level 'INVALID'.
$ zonemaster-cli --level=INVALID
--level must be one of CRITICAL, ERROR, WARNING, NOTICE, INFO, DEBUG, DEBUG2 or DEBUG3.
```

matsduf:
>     * In the synopsis, one line says `zonemaster-cli [OPTIONS] --dump-profile`. Except from `--profile=FILE`, are there any other options that could make sense when combined with `--dump-profile`? For example, `zonemaster-cli --dump-profile --no-json` still gives me JSON output because `--no-json` is not applicable to the output caused by `--dump-profile`.

I guess it would make sense to combine `--profile FILE` with `--dump-profile`.

matsduf:
@marc-vanderwal, will you create an issue or a PR of your findings?

marc-vanderwal:
> I guess it would make sense to combine `--profile FILE` with `--dump-profile`.

Yes, I agree that it makes sense; you just end up dumping the profile from `FILE`.

> @marc-vanderwal, will you create an issue or a PR of your findings?

I’ve created issue #403.

PR #374, opened 2024-06-20
Speed up Docker build

marc-vanderwal:
## Purpose

This PR amends the Dockerfile for zonemaster-cli so as to install as many prerequisite packages from binary packages as possible, thereby speeding up the build.

## Context

Release testing for v2024.1.

## Changes

 * Install most transitive dependencies for Zonemaster-CLI by means of binary packages.

## How to test this PR

Try to build the Docker image. The build process should only pick up six small packages from CPAN. Notably, Moose shouldn’t be picked up from CPAN.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,24 @@
 FROM zonemaster/engine:local as build
 
 RUN apk add --no-cache \
-    # Only needed for Net::Interface
     build-base \
+    make \
+    perl-app-cpanminus \
+    perl-cpan-meta-check \
+    perl-data-dump \
     perl-dev \
     perl-doc \
+    perl-json-xs \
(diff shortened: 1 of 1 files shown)


matsduf:
Does this increase or decrease the Docker image size?

marc-vanderwal:
There is a slight decrease. To compare:
* a `v6.1.0` image [pulled from Docker Hub](https://hub.docker.com/r/zonemaster/cli/tags) is 62.1 MB after decompression;
* currently, on `develop`, a Docker build yields an image whose size is 61.5 MB;
* with the change introduced by this PR, the image size shrinks to 60 MB.

matsduf (approved):

PR #386, opened 2024-10-09
Improved test coverage

mattias-p:
### Scope

This PR concerns itself with validation and dispatching of user-provided command line options to zonemaster-cli.

Limitations:

* The actual implementations of the options are only within scope here to the extent that proper dispatching can be determined. Any other problems with option implementations are deferred to separate issues.
* Details in how command line arguments are parsed are out of scope.
* The --progress option is out of scope.

### Goal

This PR aims to establish a sufficient regression testing suite for the stated scope. It would be great if we didn't need to perform any manual testing for this particular aspect of the application when switching out the command line parsing library.

## Context

As part of #68 we plan to drop our dependency on Moose and MooseX::Getopt::GLD. That will affect every command line option in zonemaster-cli.

## Changes

Fixes the handling of invalid --hints argument.

New tests are added for options and option combinations:

 * --restore
 * --save
 * --json-translate
 * --encoding
 * --count --raw
 * --count --json
 * --count --json-stream
 * --elapsed --raw
 * --elapsed --json
 * --elapsed --json-stream
 * --nstimes --raw
 * --nstimes --json
 * --nstimes --json-stream
 * Underscore option aliases

To help finding gaps in test coverage I added support for Devel::Cover in usage.t.

This PR also includes some cleanups.

## How to test this PR

Unit test should pass.

## Special considerations when reviewing this PR

1. Are there any options or combinations that deserve automatic tests but doesn't have any?
2. Do any options or combinations deserve more thorough automatic tests?

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -28,6 +28,11 @@ requires(
     'Zonemaster::Engine' => 6.000000,  # v6.0.0
 );
 
+test_requires(
+    'JSON::Validator'   => 0,
+    'Test::Differences' => 0,
+);
+
 # Make all platforms include inc/Module/Install/External.pm
 requires_external_bin 'find';
 if ($^O eq "freebsd") {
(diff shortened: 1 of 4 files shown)


matsduf (approved):
I see no issues with this. Nice to get better test coverage.

PR #397, opened 2024-11-22
Corrects gettext code

matsduf:
## Purpose

Some of the gettext printed messages in CLI was not correctly configured. You cannot have an Perl variable in the message. 
That has been corrected.

Without the change PO handling will not work.

## How to test this PR

* Review
* Unit test should work
* Run `zonemaster-cli` with input data that should be captured, e.g. non-existing profile file.


matsduf:
Replaced by #398

PR #416, opened 2024-12-19
Zonemaster::CLI: don’t lose early log messages

### LINKED ISSUE #402: zonemaster-cli outputs no message at all if name server given by --ns lacks required glue
On current `develop`, if `zonemaster-cli` is given fake delegation by means of `--ns`, the name server is in-bailiwick but is missing glue, then I get no error message at all:

```
$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net'
  \
Seconds Level    Testcase       Message
======= ======== ============== =======
```

Using `git bisect`, it seems that commit 69b71a6 is the one introducing the problem. If I check out its ancestor commit (2476edb), then I get the following output, which seems more correct:

```
$ zonemaster-cli --show-testcase --no-ipv6 --test basic/basic01 zonemaster.net --ns 'ns1.zonemaster.net'
Seconds Level    Testcase       Message
======= ======== ============== =======
   1.15 ERROR    Unspecified    The fake delegation of domain zonemaster.net includes an in-zone name server ns1.zonemaster.net without mandatory glue (without IP address).
   0.00 INFO     Unspecified    Using version v6.0.0 of the Zonemaster engine.
   0.00 INFO     Basic01        The zone "zonemaster.net" is found.
   0.01 INFO     Basic01        This is a test of an undelegated domain so finding the parent zone is disregarded.
   0.21 CRITICAL Basic02        There is no working name server for "zonemaster.net" so it is unreachable.
   0.21 ERROR    Basic02        Name server "ns1.zonemaster.net" cannot be resolved into an IP address.
```

marc-vanderwal:
## Purpose

This PR fixes a problem where, in one particular set of circumstances (caused by user error), `zonemaster-cli` fails to display messages on the terminal.

Adding fake delegation by means of the --ns command-line option can sometimes elicit early Zonemaster::Engine log messages that are lost because the translator isn’t properly initialized at this stage. For reasons that I do not understand fully yet, this situation causes a condition where the zonemaster-cli script displays no message at all, even though it should.

As a workaround for this problem, we set the callback to Zonemaster::Engine::Logger twice. First, we set it to a closure that simply appends them to a queue of yet-to-be-printed messages. Right after the translator is initialized, we print the header and messages stored by the first callback before switching the callback to the function we actually wanted to use, which prints the messages on the terminal.

A brief test shows no unintended side-effects with other output formats such as JSON. But the run() method in Zonemaster::CLI is long and has a high cyclomatic complexity, which doesn’t really help with testing!

## Context

Fixes #402.

## Changes

* Buffer messages generated by Zonemaster-Engine before Zonemaster-CLI is fully initialized.

## How to test this PR

Run the following command:
```
zonemaster-cli --level INFO --test basic --ns ns1.zonemaster.net zonemaster.net
```

`zonemaster-cli` should display at least one log message complaining about the fake delegation missing required glue. Previously, it would only display the table header and no message at all.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -394,86 +394,96 @@ sub run {
     my %remaining_space = ();
 
     # Callback defined here so it closes over the setup above.
-    Zonemaster::Engine->logger->callback(
-        sub {
-            my ( $entry ) = @_;
+    # But we can’t use it right now because the translator isn’t initialized.
+    my $message_printer = sub {
+        my ( $entry ) = @_;
 
-            print_spinner() if $show_progress;
+        print_spinner() if $show_progress;
 
-            $counter{ uc $entry->level } += 1;
+        $counter{ uc $entry->level } += 1;
 
-            if ( $numeric{ uc $entry->level } >= $numeric{$opt_level} ) {
-                $printed_something = 1;
+        if ( $numeric{ uc $entry->level } >= $numeric{$opt_level} ) {
+            $printed_something = 1;
 
-                if ( $opt_json and $opt_json_stream ) {
-                    my %r;
+            if ( $opt_json and $opt_json_stream ) {
+                my %r;
 
-                    $r{timestamp} = $entry->timestamp if $opt_time;
-                    $r{module}    = $entry->module    if $opt_show_module;
-                    $r{testcase}  = $entry->testcase  if $opt_show_testcase;
-                    $r{tag}       = $entry->tag;
-                    $r{level}     = $entry->level if $opt_show_level;
-                    $r{args}      = $entry->args  if $entry->args;
-                    $r{message}   = $translator->translate_tag( $entry ) unless $opt_raw;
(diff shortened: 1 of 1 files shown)


matsduf:
> Run the following command:
> 
> ```
> zonemaster-cli --level INFO --test basic01 --ns ns1.zonemaster.net zonemaster.net
> ```
> 
> `zonemaster-cli` should display at least one log message complaining about the fake delegation missing required glue. Previously, it would only display the table header and no message at all.

It should display messages from Basic01 too. I think it is better to run "--test Basic" to show that it shows messages from both Basic01 and Basic02.

marc-vanderwal:
> It should display messages from Basic01 too. I think it is better to run "--test Basic" to show that it shows messages from both Basic01 and Basic02.

Good idea; I’ve updated the testing procedure in the PR text.

mattias-p (approved):

PR #419, opened 2025-01-27
Slow down the spinner a bit

marc-vanderwal:
## Purpose

This PR slows down the spinner in zonemaster-cli to improve the animation’s appearance on the terminal and to avoid overloading the terminal’s standard output with thousands of calls.

In a nutshell, the code no longer advances the animation if the current frame has been showing for less than 0,1 seconds.

## Context

Searching for performance improvements to be made in `zonemaster-cli`.

## Changes

Slow down the spinner.

## How to test this PR

Run `zonemaster-cli zonemaster.net`. The spinner’s animation should be slower, and actually look like it’s spinning. The animation may still pause briefly a few times, like before: that’s because Zonemaster-Engine sometimes does not emit any message for a few seconds while it is waiting for a response from remote name servers.

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -26,6 +26,7 @@ use Pod::Usage;
 use POSIX qw[setlocale LC_MESSAGES LC_CTYPE];
 use Readonly;
 use Scalar::Util qw[blessed];
+use Time::HiRes;
 use Try::Tiny;
 use Zonemaster::LDNS;
 use Zonemaster::Engine;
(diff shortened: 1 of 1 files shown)


tgreenx (dismissed):

mattias-p (dismissed):

marc-vanderwal:
Rebased on latest `develop` in order to address conflicts. No functional change compared to the previous iteration.

@tgreenx and @mattias-p, can I have another review, please?

mattias-p (approved):

tgreenx (approved):

PR #421, opened 2025-02-11
Expand --nstimes option

tgreenx:
## Purpose

This PR proposes to expand the `--nstimes` option in several ways. See the "Changes" section below.

Note: this option applies only to "sent" queries. Cached queries are not counted.

## Context

N/A

## Changes

- Extend to all queried name servers: this means that non-queried name servers will not appear, with the exception of name servers of the child or parent zone
- Add classification of name servers ("child", "parent", "other"). Note that name servers shared by the child and parent zone will appear in both categories.
- Add "Count" column as the number of queries sent per name server
- Add "Grand total" row for query times and query count
- Refactoring

## How to test this PR

Run any test with the `--nstimes` option. Check that the formatting is done correctly.

Example output:
```
$ zonemaster-cli --nstimes --raw --no-ipv6 --test basic02 afnic.fr

                     Name servers         Max        Min        Avg     Stddev     Median       Total       Count
=================================  ========== ========== ========== ========== ========== =========== ===========
Child zone ----------------------
          g.ext.nic.fr/194.0.36.1      110.98      33.27      72.88      22.62      69.07     1093.16          15
      g.ext.nic.fr/2001:678:4c::1        0.00       0.00       0.00       0.00       0.00        0.00           0
           ns1.nic.fr/192.134.4.1        4.78       4.78       4.78       0.00       4.78        4.78           1
  ns1.nic.fr/2001:67c:2218:2::4:1        0.00       0.00       0.00       0.00       0.00        0.00           0
            ns2.nic.fr/192.93.0.4        4.82       4.82       4.82       0.00       4.82        4.82           1
  ns2.nic.fr/2001:660:3005:1::1:2        0.00       0.00       0.00       0.00       0.00        0.00           0
          ns3.nic.fr/192.134.0.49        4.96       4.96       4.96       0.00       4.96        4.96           1
  ns3.nic.fr/2001:660:3006:1::1:1        0.00       0.00       0.00       0.00       0.00        0.00           0
Parent zone ---------------------
               d.nic.fr/194.0.9.1      113.08      60.96      87.02      26.06      87.02      174.05           2
           d.nic.fr/2001:678:c::1        0.00       0.00       0.00       0.00       0.00        0.00           0
      f.ext.nic.fr/194.146.106.46        0.00       0.00       0.00       0.00       0.00        0.00           0
f.ext.nic.fr/2001:67c:1010:11::53        0.00       0.00       0.00       0.00       0.00        0.00           0
          g.ext.nic.fr/194.0.36.1      110.98      33.27      72.88      22.62      69.07     1093.16          15
      g.ext.nic.fr/2001:678:4c::1        0.00       0.00       0.00       0.00       0.00        0.00           0
Other ---------------------------
    a.root-servers.net/198.41.0.4      113.63     113.63     113.63       0.00     113.63      113.63           1
  m.root-servers.net/202.12.27.33      110.43       8.39      43.07      21.62      39.96      602.93          14
=================================  ========== ========== ========== ========== ========== =========== ===========
                      Grand total                                                             1998.33          35
```

### DIFF
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -9,7 +9,6 @@ package Zonemaster::CLI;
 
 use v5.26;
 
-use strict;
 use warnings;
 
 use version; our $VERSION = version->declare( "v7.2.0" );
@@ -646,17 +645,16 @@ sub run {
         my %all_nss = %{ Zonemaster::Engine::Nameserver::object_cache };
         my @child_nss = @{ $zone->ns };
         my @parent_nss = @{ $zone->parent->ns };
-        my @nss;
+        my @all_responded_nss;
 
         foreach my $ns_name ( keys %all_nss ) {
             foreach my $ns ( values %{ $all_nss{$ns_name} } ) {
-                push @nss, $ns if scalar @{ $ns->times } > 0;
+                push @all_responded_nss, $ns if scalar @{ $ns->times } > 0;
             }
         }
 
-        my %nss_filter;
-        @nss_filter{ ( @child_nss, @parent_nss ) } = undef;
-        my @other_nss = grep { ! exists $nss_filter{$_} } @nss;
+        my %nss_filter = map { $_ => undef } ( @child_nss, @parent_nss );
+        my @other_nss = grep { ! exists $nss_filter{$_} } @all_responded_nss;
 
         if ( $opt_json ) {
             my @times;
@@ -690,7 +688,7 @@ sub run {
         }
         else {
             my $header = __( 'Name servers' );
-            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @nss ), $header );
+            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @all_responded_nss ), $header );
             printf "\n%${max}s %s\n", $header, '        Max        Min        Avg     Stddev     Median       Total       Count';
             printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
 
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -644,22 +644,22 @@ sub run {
     if ( $opt_nstimes ) {
         my $zone = Zonemaster::Engine->zone( $domain );
         my %all_nss = %{ Zonemaster::Engine::Nameserver::object_cache };
-        my @zone_nss = @{ $zone->ns };
+        my @child_nss = @{ $zone->ns };
         my @parent_nss = @{ $zone->parent->ns };
         my @nss;
 
-        foreach my $keys ( keys %all_nss ) {
-            foreach my $val ( values %{ $all_nss{$keys} } ) {
-                push @nss, $val if scalar @{ $val->times } > 0;
+        foreach my $ns_name ( keys %all_nss ) {
+            foreach my $ns ( values %{ $all_nss{$ns_name} } ) {
+                push @nss, $ns if scalar @{ $ns->times } > 0;
             }
         }
 
         my %nss_filter;
-        @nss_filter{ ( @zone_nss, @parent_nss ) } = ();
-        my @filtered_nss = grep { ! exists $nss_filter{$_} } @nss;
+        @nss_filter{ ( @child_nss, @parent_nss ) } = undef;
+        my @other_nss = grep { ! exists $nss_filter{$_} } @nss;
 
         if ( $opt_json ) {
-            my ( @times, @items ) = ();
+            my @times;
 
             sub json_nstimes {
                 my ( $ns ) = @_;
@@ -674,75 +674,71 @@ sub run {
                     'count' => scalar @{ $ns->times } };
             }
 
-            foreach my $ns ( sort @zone_nss ) {
-                push @items, json_nstimes( $ns );
+            if ( @child_nss ) {
+                my @entries = map { json_nstimes( $_ ) } sort @child_nss;
+                push @times, { 'zone' => \@entries };
             }
-            push @times, { 'zone' => \@items } if scalar @items;
-            @items = ();
 
-            foreach my $ns ( sort @parent_nss ) {
-                push @items, json_nstimes( $ns );
+            if ( @parent_nss ) {
+                my @entries = map { json_nstimes( $_ ) } sort @parent_nss;
+                push @times, { 'parent' => \@entries };
             }
-            push @times, { 'parent' => \@items } if scalar @items;
-            @items = ();
 
-            foreach my $ns ( sort @filtered_nss ) {
-                push @items, json_nstimes( $ns );
+            if ( @other_nss ) {
+                my @entries = map { json_nstimes( $_ ) } sort @other_nss;
+                push @times, { 'other' => \@entries };
             }
-            push @times, { 'other' => \@items } if scalar @items;
 
             $json_output->{nstimes} = \@times;
         }
         else {
-            my $max = max map { length( "$_" ) } ( uniq ( @zone_nss, @parent_nss, @nss ), q{Server} );
+            my $header = __( 'Name servers' );
+            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @nss ), $header );
             print "\n";
-            printf __("%${max}s %s\n"), 'Server', '      Max      Min      Avg   Stddev   Median     Total     Count';
-            printf "%${max}s %s\n", '=' x $max, ' ======== ======== ======== ======== ======== ========= =========';
+            printf "%${max}s %s\n", $header, '        Max        Min        Avg     Stddev     Median       Total       Count';
+            printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
 
             my $total_queries_count = 0;
             my $total_queries_times = 0;
             my %nss_already_processed;
 
             sub print_nstimes {
-                my ( $ns, $max, $total_queries_count, $total_queries_times, %nss_already_processed ) = @_;
+                my ( $ns, $max, $total_queries_count, $total_queries_times, $nss_already_processed_ref ) = @_;
+                my %nss_already_processed = %{ $nss_already_processed_ref };
 
                 printf "%${max}s ", $ns->string;
-                printf "%9.2f ",    1000 * $ns->max_time;
-                printf "%8.2f ",    1000 * $ns->min_time;
-                printf "%8.2f ",    1000 * $ns->average_time;
-                printf "%8.2f ",    1000 * $ns->stddev_time;
-                printf "%8.2f ",    1000 * $ns->median_time;
-                printf "%9.2f ",    1000 * $ns->sum_time;
-                printf "%9d\n",     scalar @{ $ns->times };
+                printf "%11.2f ",    1000 * $ns->max_time;
+                printf "%10.2f ",    1000 * $ns->min_time;
+                printf "%10.2f ",    1000 * $ns->average_time;
+                printf "%10.2f ",    1000 * $ns->stddev_time;
+                printf "%10.2f ",    1000 * $ns->median_time;
+                printf "%11.2f ",    1000 * $ns->sum_time;
+                printf "%11d\n",     scalar @{ $ns->times };
                 $total_queries_count += scalar @{ $ns->times } unless $nss_already_processed{$ns};
                 $total_queries_times += ( 1000 * $ns->sum_time ) unless $nss_already_processed{$ns};
 
-                return $total_queries_count, $total_queries_times, %nss_already_processed;
+                return $total_queries_count, $total_queries_times;
             }
 
-            printf __("%s %s\n"), 'Child', '-' x ( ( $max - length 'Child' ) - 1 );
-            foreach my $ns ( sort @zone_nss ) {
-                ( $total_queries_count, $total_queries_times, %nss_already_processed ) =
-                    print_nstimes( $ns, $max, $total_queries_count, $total_queries_times, %nss_already_processed );
-                $nss_already_processed{$ns} = 1;
+            my %section_mapping = (
+                1 => { __( 'Child zone' ) => \@child_nss },
+                2 => { __( 'Parent zone' ) => \@parent_nss },
+                3 => { __( 'Other' ) => \@other_nss }
+            );
+
+            foreach my $section_order ( sort keys %section_mapping ) {
+                foreach my $section_header ( keys % { $section_mapping{$section_order} } ) {
+                    printf "%s %s\n", $section_header, '-' x ( ( $max - length $section_header ) - 1 );
+                    foreach my $section_nss ( sort @{ $section_mapping{$section_order}{$section_header} } ) {
+                        ( $total_queries_count, $total_queries_times ) =
+                            print_nstimes( $section_nss, $max, $total_queries_count, $total_queries_times, \%nss_already_processed );
+                        $nss_already_processed{$section_nss} = 1;
+                    }
+                }
             }
 
-            printf __("%s %s\n"), 'Parent', '-' x ( ( $max - length 'Parent' ) - 1 );
-            foreach my $ns ( sort @parent_nss ) {
-                ( $total_queries_count, $total_queries_times, %nss_already_processed ) =
-                    print_nstimes( $ns, $max, $total_queries_count, $total_queries_times, %nss_already_processed );
-                $nss_already_processed{$ns} = 1;
-            }
-
-            printf __("%s %s\n"), 'Other', '-' x ( ( $max - length 'Other' ) - 1 );
-            foreach my $ns ( sort @filtered_nss ) {
-                ( $total_queries_count, $total_queries_times, %nss_already_processed ) =
-                    print_nstimes( $ns, $max, $total_queries_count, $total_queries_times, %nss_already_processed );
-                $nss_already_processed{$ns} = 1;
-            }
-
-            printf "%${max}s %s\n", '=' x $max, ' ======== ======== ======== ======== ======== ========= =========';
-            printf __("%${max}s %55.2f %9s\n"), 'Total', $total_queries_times, $total_queries_count;
+            printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
+            printf "%${max}s %67.2f %11s\n", __( 'Grand total' ), $total_queries_times, $total_queries_count;
         }
     } ## end if ( $opt_nstimes )
 
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -1,13 +1,13 @@
 # Brief help module to define the exception we use for early exits.
 package Zonemaster::Engine::Exception::NormalExit;
-use 5.014002;
+use v5.26;
 use warnings;
 use parent 'Zonemaster::Engine::Exception';
 
 # The actual interesting module.
 package Zonemaster::CLI;
 
-use 5.014002;
+use v5.26;
 
 use strict;
 use warnings;
@@ -661,7 +661,7 @@ sub run {
         if ( $opt_json ) {
             my @times;
 
-            sub json_nstimes {
+            my sub json_nstimes {
                 my ( $ns ) = @_;
                 return {
                     'ns'      => $ns->string,
@@ -698,7 +698,7 @@ sub run {
             my $total_queries_times = 0;
             my %nss_already_processed;
 
-            sub print_nstimes {
+            my sub print_nstimes {
                 my ( $ns, $max, $total_queries_count, $total_queries_times, $nss_already_processed_ref ) = @_;
                 my %nss_already_processed = %{ $nss_already_processed_ref };
 
--- a/lib/Zonemaster/CLI.pm
+++ b/lib/Zonemaster/CLI.pm
@@ -643,38 +643,106 @@ sub run {
 
     if ( $opt_nstimes ) {
         my $zone = Zonemaster::Engine->zone( $domain );
-        my $max  = max map { length( "$_" ) } ( @{ $zone->ns }, q{Server} );
+        my %all_nss = %{ Zonemaster::Engine::Nameserver::object_cache };
+        my @zone_nss = @{ $zone->ns };
+        my @parent_nss = @{ $zone->parent->ns };
+        my @nss;
+
+        foreach my $keys ( keys %all_nss ) {
+            foreach my $val ( values %{ $all_nss{$keys} } ) {
+                push @nss, $val if scalar @{ $val->times } > 0;
+            }
+        }
+
+        my %nss_filter;
+        @nss_filter{ ( @zone_nss, @parent_nss ) } = ();
+        my @filtered_nss = grep { ! exists $nss_filter{$_} } @nss;
 
         if ( $opt_json ) {
-            my @times = ();
-            foreach my $ns ( @{ $zone->ns } ) {
-                push @times,
-                  {
-                    'ns'     => $ns->string,
-                    'max'    => 1000 * $ns->max_time,
-                    'min'    => 1000 * $ns->min_time,
-                    'avg'    => 1000 * $ns->average_time,
-                    'stddev' => 1000 * $ns->stddev_time,
-                    'median' => 1000 * $ns->median_time,
-                    'total'  => 1000 * $ns->sum_time
-                  };
+            my ( @times, @items ) = ();
+
+            sub json_nstimes {
+                my ( $ns ) = @_;
+                return {
+                    'ns'      => $ns->string,
+                    'max'     => 1000 * $ns->max_time,
+                    'min'     => 1000 * $ns->min_time,
+                    'avg'     => 1000 * $ns->average_time,
+                    'stddev'  => 1000 * $ns->stddev_time,
+                    'median'  => 1000 * $ns->median_time,
+                    'total'   => 1000 * $ns->sum_time,
+                    'count' => scalar @{ $ns->times } };
             }
+
+            foreach my $ns ( sort @zone_nss ) {
+                push @items, json_nstimes( $ns );
+            }
+            push @times, { 'zone' => \@items } if scalar @items;
+            @items = ();
+
+            foreach my $ns ( sort @parent_nss ) {
+                push @items, json_nstimes( $ns );
+            }
+            push @times, { 'parent' => \@items } if scalar @items;
+            @items = ();
+
+            foreach my $ns ( sort @filtered_nss ) {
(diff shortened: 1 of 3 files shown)


mattias-p (dismissed):
This change makes the `--nstimes` option go outside of timing territory and into the wider statistics territory, I guess. I think it begs the question if there are other per-nameserver stats that we would be interested in, and if so we should consider having an `--nsstats` option. But IMHO that's out of scope for this PR.

marc-vanderwal (dismissed):
Yes, for those statistics it makes perfect sense to have the sample size. Could we also have a count of the number of queries answered? (Or conversely, the number of “lost” packets?)

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
+        my @nss;
+
+        foreach my $keys ( keys %all_nss ) {
+            foreach my $val ( values %{ $all_nss{$keys} } ) {
+                push @nss, $val if scalar @{ $val->times } > 0;
+            }
+        }
+
+        my %nss_filter;
+        @nss_filter{ ( @zone_nss, @parent_nss ) } = ();
```
mattias-p: This gives the same result but is more straight forward.

```suggestion
        @nss_filter{ ( @zone_nss, @parent_nss ) } = undef;
```

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
-                push @times, {
-                    'ns'     => $ns->string,
-                    'max'    => 1000 * $ns->max_time,
-                    'min'    => 1000 * $ns->min_time,
-                    'avg'    => 1000 * $ns->average_time,
-                    'stddev' => 1000 * $ns->stddev_time,
-                    'median' => 1000 * $ns->median_time,
-                    'total'  => 1000 * $ns->sum_time
-                };
+            my ( @times, @items ) = ();
```
mattias-p: This gives the same result but is more straight forward.

```suggestion
            my ( @times, @items );
```

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
+                    'median'  => 1000 * $ns->median_time,
+                    'total'   => 1000 * $ns->sum_time,
+                    'count' => scalar @{ $ns->times } };
+            }
+
+            foreach my $ns ( sort @zone_nss ) {
+                push @items, json_nstimes( $ns );
             }
+            push @times, { 'zone' => \@items } if scalar @items;
+            @items = ();
```
mattias-p: Does this really work? I mean, references are created to the same array on line 706, 712 and 718.

I suggest declaring a new array for each @times entry. Also, "times" and "items" are visually very similar, so I renamed one of them in my suggestion.

```suggestion
            if ( @zone_nss ) {
                my @entries = map { json_times( $_ ) } sort @zone_nss;
                push @times, { zone => \@entries };
            }
```

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
+
+            foreach my $ns ( sort @filtered_nss ) {
+                push @items, json_nstimes( $ns );
+            }
+            push @times, { 'other' => \@items } if scalar @items;
+
             $json_output->{nstimes} = \@times;
         }
         else {
+            my $max = max map { length( "$_" ) } ( uniq ( @zone_nss, @parent_nss, @nss ), q{Server} );
```
mattias-p: Isn't it both simpler and cheaper to just run `length` on everything instead of deduplicating the entries with `uniq`?

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
             print "\n";
-            printf "%${max}s %s\n", 'Server', '      Max      Min      Avg   Stddev   Median     Total';
-            printf "%${max}s %s\n", '=' x $max, ' ======== ======== ======== ======== ======== =========';
+            printf __("%${max}s %s\n"), 'Server', '      Max      Min      Avg   Stddev   Median     Total     Count';
```
mattias-p: The call to `__()` makes no sense here. First, the argument to `__()` must be a string literal without any string interpolation like "$foo". Second, the argument to `__()` should contain translatable English text, this is just a format string.

There are more of these.

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
+            printf "%${max}s %s\n", '=' x $max, ' ======== ======== ======== ======== ======== ========= =========';
+
+            my $total_queries_count = 0;
+            my $total_queries_times = 0;
+            my %nss_already_processed;
+
+            sub print_nstimes {
```
mattias-p: I tried to look up the exact semantics of defining a subroutine within another subroutine like this, but I didn't find anything conclusive on the subject. Perl 5.26 stabilized a syntax for lexical subroutines (i.e. `my sub`). As it happens this is the newest Perl version we could depend on, so I guess we could add `use v5.26;` and use the `my sub` syntax here. Other possibilities would be to use the old-school `my $print_nstimes = sub { ... };` syntax, or just move it out into the top level of the file.

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,106 @@ sub run {
...
                 printf "%8.2f ",    1000 * $ns->average_time;
                 printf "%8.2f ",    1000 * $ns->stddev_time;
                 printf "%8.2f ",    1000 * $ns->median_time;
-                printf "%9.2f\n",   1000 * $ns->sum_time;
+                printf "%9.2f ",    1000 * $ns->sum_time;
+                printf "%9d\n",     scalar @{ $ns->times };
+                $total_queries_count += scalar @{ $ns->times } unless $nss_already_processed{$ns};
+                $total_queries_times += ( 1000 * $ns->sum_time ) unless $nss_already_processed{$ns};
+
+                return $total_queries_count, $total_queries_times, %nss_already_processed;
```
mattias-p: Isn't it better to move this summation out from print_nstimes and into its own loop?

matsduf:
What are the use cases for using `--nstimes`? 

> Note: this option applies only to "sent" queries. Cached queries are not counted.

Is there any technical reason to exclude those? Otherwise it could be reasonable to see the queries to cache too. Will there be any difference if global or local cache is used? If so, it can be confusing when using global cache.

mattias-p:
> > Note: this option applies only to "sent" queries. Cached queries are not counted.
> 
> Is there any technical reason to exclude those? Otherwise it could be reasonable to see the queries to cache too.

AFAICT what is actually interesting is the number of distinct requests. That number should to be the same whether or not recorded data is used, and whether or not intermediate results within Engine are cached. I believe the number of queries that are actually sent, the number of queries that would be sent if we didn't cache, and the sum of those are of little interest.

script/zonemaster-cli
```diff
@@ -255,8 +255,8 @@ level that were logged during the run.
 
 =item B<--[no-]nstimes>
 
-Print a summary, at the end of a run, of the times (in milliseconds) the zone's
-name servers took to answer.
+Print a summary, at the end of a run, of the times (in milliseconds) the name servers
+took to answer, as well as the number of sent queries.
```
matsduf: As I understand it, the time each name server takes it the total time over all queries.

```suggestion
Print a summary, at the end of a run, of the total time (in milliseconds) each name server
took to answer all queries, as well as the number of sent queries per name server.
```

matsduf:
> AFAICT what is actually interesting is the number of distinct requests. That number should to be the same whether or not recorded data is used, and whether or not intermediate results within Engine are cached. I believe the number of queries that are actually sent, the number of queries that would be sent if we didn't cache, and the sum of those are of little interest.

@mattias-p, what use case do consider when writing this? Do you see other use cases of the `--ns-times` option?

mattias-p:
> > AFAICT what is actually interesting is the number of distinct requests. That number should to be the same whether or not recorded data is used, and whether or not intermediate results within Engine are cached. I believe the number of queries that are actually sent, the number of queries that would be sent if we didn't cache, and the sum of those are of little interest.
> 
> @mattias-p, what use case do consider when writing this? Do you see other use cases of the `--ns-times` option?

I'm not really thinking about use cases. I'm thinking about how to make meaningful measurements. A measurement is more meaningful when it is variable in a single dimension and isn't influenced by multiple different things. So I meant to suggest that we could define the count so that it isn't affected by whether caching options and implementation details.

matsduf:
I think it is reasonable to have some use cases in mind when defining what figures to extract and present. Maybe it could be useful to distinguish between queries that are responded from cache and those that require sending a message externally.

lib/Zonemaster/CLI.pm
tgreenx: Fixed

lib/Zonemaster/CLI.pm
tgreenx: Fixed

lib/Zonemaster/CLI.pm
tgreenx: Fixed

lib/Zonemaster/CLI.pm
tgreenx: Sure, fixed

lib/Zonemaster/CLI.pm
tgreenx: You're right. Fixed (in all places).

script/zonemaster-cli
tgreenx: I'm not sure this is an improvement as is. It doesn't just print the total time, as there is also e.g. average, median times, etc.
Instead, rephrased to:

> Print a summary, at the end of a run, of various metrics related to the times (in milliseconds) each name server took to answer, as well as a count of the number of sent queries per name server.

tgreenx:
@mattias-p @matsduf I should have addressed most comments, and I've also made small refactoring/improvements, see https://github.com/zonemaster/zonemaster-cli/pull/421/commits/b0c1bf2540dc2be2cee08e9cbc73c2d339f9b237. Please re-review.

And regarding the use cases, as Mattias pointed out, the purpose of this option is to measure response times of name servers, so that automatically excludes cached queries.

tgreenx:
@mattias-p @matsduf @marc-vanderwal please (re-)review.

lib/Zonemaster/CLI.pm
marc-vanderwal: Yes. Nested `sub`s have never been officially supported in Perl. Either use [`my sub`](https://perldoc.perl.org/5.26.0/perlsub#Lexical-Subroutines), introduced in Perl 5.26, or move it to the top level of the file.

lib/Zonemaster/CLI.pm
```diff
@@ -669,37 +669,99 @@ sub run {
...
+                    'ns'      => $ns->string,
+                    'max'     => 1000 * $ns->max_time,
+                    'min'     => 1000 * $ns->min_time,
+                    'avg'     => 1000 * $ns->average_time,
+                    'stddev'  => 1000 * $ns->stddev_time,
+                    'median'  => 1000 * $ns->median_time,
+                    'total'   => 1000 * $ns->sum_time,
+                    'count'   => scalar @{ $ns->times }
                 };
             }
```
marc-vanderwal: Nested `sub`s aren’t officially supported in Perl. Either use [`my sub`](https://perldoc.perl.org/5.26.0/perlsub#Lexical-Subroutines) with `use v5.26` or move it to the top-level of the file.

marc-vanderwal (commented):
Looks fine, though there’s one thing left to address before I can approve.

lib/Zonemaster/CLI.pm
tgreenx: Done

lib/Zonemaster/CLI.pm
tgreenx: Done

tgreenx:
@matsduf @mattias-p @marc-vanderwal Rebased on latest develop, and addressed review comments (see https://github.com/zonemaster/zonemaster-cli/pull/421/commits/acf65f511822f78d551d2ba82d8dd753790c5ba4). Please re-review.

lib/Zonemaster/CLI.pm
```diff
@@ -1,13 +1,13 @@
...
-use 5.014002;
+use v5.26;
 use warnings;
 use parent 'Zonemaster::Engine::Exception';
 
 # The actual interesting module.
 package Zonemaster::CLI;
 
-use 5.014002;
+use v5.26;
```
mattias-p, line 10: Maybe also remove `use strict` since that's implicit in `use v5.26`.

tgreenx:
@matsduf @mattias-p @marc-vanderwal please re-review this PR.

matsduf (commented):
This looks fine as function, but I think there is a lack of documentation. I am not sure that the best place is to add documentation into the scripts. Rather I think a markdown document in the document tree would be better, but a reference from here.

Please share your thoughts of documentation and I am ready to approve.

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +682,99 @@ sub run {
 
     if ( $opt_nstimes ) {
         my $zone = Zonemaster::Engine->zone( $domain );
-        my $max = max map { length( "$_" ) } ( @{ $zone->ns }, q{Server} );
+        my %all_nss = %{ Zonemaster::Engine::Nameserver::object_cache };
+        my @child_nss = @{ $zone->ns };
+        my @parent_nss = @{ $zone->parent->ns };
+        my @nss;
```
mattias-p: Maybe give this one a more descriptive name, e.g., `@all_responding_nss`?

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +682,99 @@ sub run {
...
+
+        foreach my $ns_name ( keys %all_nss ) {
+            foreach my $ns ( values %{ $all_nss{$ns_name} } ) {
+                push @nss, $ns if scalar @{ $ns->times } > 0;
+            }
+        }
+
+        my %nss_filter;
+        @nss_filter{ ( @child_nss, @parent_nss ) } = undef;
```
mattias-p: How would you feel about doing it this way instead?

```suggestion
        my %nss_filter = map { $_ => undef } @child_nss, @parent_nss;
```

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +682,99 @@ sub run {
...
-            foreach my $ns ( @{ $zone->ns } ) {
-                printf "%${max}s ", $ns->string;
-                printf "%9.2f ",    1000 * $ns->max_time;
-                printf "%8.2f ",    1000 * $ns->min_time;
-                printf "%8.2f ",    1000 * $ns->average_time;
-                printf "%8.2f ",    1000 * $ns->stddev_time;
-                printf "%8.2f ",    1000 * $ns->median_time;
-                printf "%9.2f\n",   1000 * $ns->sum_time;
+            my $header = __( 'Name servers' );
+            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @nss ), $header );
```
mattias-p: Maybe it's just me not understanding this, but aren't child_nss and parent_nss subsets of nss?

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +682,99 @@ sub run {
...
+                    'median'  => 1000 * $ns->median_time,
+                    'total'   => 1000 * $ns->sum_time,
+                    'count'   => scalar @{ $ns->times }
                 };
             }
+
+            my %section_mapping = (
+                'child' => \@child_nss,
+                'parent' => \@parent_nss,
+                'other' => \@other_nss
```
mattias-p, line 679: If I'm reading this correctly, child and parent include both responding and non-responding name servers, but other only includes non-responding ones. Is that the intention?

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +682,99 @@ sub run {
...
+                    foreach my $section_nss ( sort @{ $section_mapping{$section_order}{$section_header} } ) {
+                        ( $total_queries_count, $total_queries_times ) =
+                            print_nstimes( $section_nss, $max, $total_queries_count, $total_queries_times, \%nss_already_processed );
+                        $nss_already_processed{$section_nss} = 1;
+                    }
+                }
+            }
+
+            printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
+            printf "%${max}s %67.2f %11s\n", __( 'Grand total' ), $total_queries_times, $total_queries_count;
```
mattias-p, line 736: Shouldn't the grand total be included in the JSON output also?

mattias-p (commented):
I comments for a bunch for remarks and questions and whatnot. And here's another one that I didn't know where to put as a comment:

The section mapping is defined once for the JSON output and once for table output. And the grand totals are calculated as part of the table output logic. IMHO it would be better to set up the data to be shown just once and do it before we branch to the individual output formats.

lib/Zonemaster/CLI.pm
tgreenx: Yeah, better.

lib/Zonemaster/CLI.pm
tgreenx: Done

lib/Zonemaster/CLI.pm
tgreenx: Yeah it is, so as to shorten the output while retaining meaningful information for the zone.

lib/Zonemaster/CLI.pm
tgreenx: Yes, but instead changed to `all_responded_nss` to clarify further.

lib/Zonemaster/CLI.pm
tgreenx: No because `nss` contains only name servers which responded, while `{child, parent}_nss` could have name servers that didn't get queried.

lib/Zonemaster/CLI.pm
tgreenx: I wondered about that too, but decided against it for now. It could be done more easily once I do the refactoring you suggested.

tgreenx (commented):
> This looks fine as function, but I think there is a lack of documentation. I am not sure that the best place is to add documentation into the scripts. Rather I think a markdown document in the document tree would be better, but a reference from here.
> 
> Please share your thoughts of documentation and I am ready to approve.

So far I think all the options are described that way. I could attempt to extend the documentation of this option in a future PR, although I don't immediately see what is missing. I think the current option description and its output are rather self-explanatory, considering that I've added explicit headers to the output.

> The section mapping is defined once for the JSON output and once for table output. And the grand totals are calculated as part of the table output logic. IMHO it would be better to set up the data to be shown just once and do it before we branch to the individual output formats.

Yes you're right, if you don't mind though I'd prefer to have it done in a future PR (I can open an issue to track it) so that this one can be merged in time for v2025.1. Is that fine with you?

matsduf:
> > This looks fine as function, but I think there is a lack of documentation. I am not sure that the best place is to add documentation into the scripts. Rather I think a markdown document in the document tree would be better, but a reference from here.
> > Please share your thoughts of documentation and I am ready to approve.
> 
> So far I think all the options are described that way. I could attempt to extend the documentation of this option in a future PR, although I don't immediately see what is missing. I think the current option description and its output are rather self-explanatory, considering that I've added explicit headers to the output.

Please look at it from a "foreign" eye, and then the output is far from self-explanatory.

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +681,98 @@ sub run {
...
-                printf "%9.2f ",    1000 * $ns->max_time;
-                printf "%8.2f ",    1000 * $ns->min_time;
-                printf "%8.2f ",    1000 * $ns->average_time;
-                printf "%8.2f ",    1000 * $ns->stddev_time;
-                printf "%8.2f ",    1000 * $ns->median_time;
-                printf "%9.2f\n",   1000 * $ns->sum_time;
+            my $header = __( 'Name servers' );
+            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @all_responded_nss ), $header );
+            printf "\n%${max}s %s\n", $header, '        Max        Min        Avg     Stddev     Median       Total       Count';
+            printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
```
marc-vanderwal, line 693: This table’s columns aren’t fully localized and that’s a problem.

Also, it's better to write `printf "%*s", $width, $string` instead of `printf "%${width}s", $string`.

```suggestion
            my $max = max map { length( "$_" ) } ( ( @child_nss, @parent_nss, @all_responded_nss ), __( 'Name servers' ) );
            my @columns = (
                { label => __( 'Name servers' ), width => $max },
                { label => __( 'Max' ),          width => 11 },
                { label => __( 'Min' ),          width => 10 },
                { label => __( 'Avg' ),          width => 10 },
                { label => __( 'Stddev' ),       width => 10 },
                { label => __( 'Median' ),       width => 10 },
                { label => __( 'Total' ),        width => 11 },
                { label => __( 'Count' ),        width => 11 },
            );

            # Table header
            print "\n";
            say join " ", map { sprintf "%*s", $_->{width}, $_->{label} } @columns;
            say join " ", map { "=" x $_->{width} } @columns;
```

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +681,98 @@ sub run {
...
+                my %nss_already_processed = %{ $nss_already_processed_ref };
+
+                printf "%${max}s ",  $ns->string;
+                printf "%11.2f ",    1000 * $ns->max_time;
+                printf "%10.2f ",    1000 * $ns->min_time;
+                printf "%10.2f ",    1000 * $ns->average_time;
+                printf "%10.2f ",    1000 * $ns->stddev_time;
+                printf "%10.2f ",    1000 * $ns->median_time;
+                printf "%11.2f ",    1000 * $ns->sum_time;
+                printf "%11d\n",     scalar @{ $ns->times };
```
marc-vanderwal, line 710: Thanks to introducing the `@columns` variable above in my previous suggestion, we can have a single source of truth for the column widths.

And there’s another trick: by using `%*2$s` instead of `%*s` as a format string, the width field can be put second instead of first in the call to `printf`.
```suggestion
                printf '%*2$s ',     $ns->string,              $columns[0]->{width};
                printf '%*2$.2f ',   1000 * $ns->max_time,     $columns[1]->{width};
                printf '%*2$.2f ',   1000 * $ns->min_time,     $columns[2]->{width};
                printf '%*2$.2f ',   1000 * $ns->average_time, $columns[3]->{width};
                printf '%*2$.2f ',   1000 * $ns->stddev_time,  $columns[4]->{width};
                printf '%*2$.2f ',   1000 * $ns->median_time,  $columns[5]->{width};
                printf '%*2$.2f ',   1000 * $ns->sum_time,     $columns[6]->{width};
                printf "%*2\$d\n",   scalar @{ $ns->times },   $columns[7]->{width};
```

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +681,98 @@ sub run {
...
+
+            my %section_mapping = (
+                1 => { __( 'Child zone' ) => \@child_nss },
+                2 => { __( 'Parent zone' ) => \@parent_nss },
+                3 => { __( 'Other' ) => \@other_nss }
+            );
+
+            foreach my $section_order ( sort keys %section_mapping ) {
+                foreach my $section_header ( keys % { $section_mapping{$section_order} } ) {
+                    printf "%s %s\n", $section_header, '-' x ( ( $max - length $section_header ) - 1 );
```
marc-vanderwal, line 725: Using `printf` to do such a thing is discouraged:
```suggestion
                    say $section_header, ' ', '-' x ( ( $max - length $section_header ) - 1 );
```

lib/Zonemaster/CLI.pm
```diff
@@ -682,37 +681,98 @@ sub run {
...
+                    foreach my $section_nss ( sort @{ $section_mapping{$section_order}{$section_header} } ) {
+                        ( $total_queries_count, $total_queries_times ) =
+                            print_nstimes( $section_nss, $max, $total_queries_count, $total_queries_times, \%nss_already_processed );
+                        $nss_already_processed{$section_nss} = 1;
+                    }
+                }
+            }
+
+            printf "%${max}s %s\n", '=' x $max, ' ========== ========== ========== ========== ========== =========== ===========';
+            printf "%${max}s %67.2f %11s\n", __( 'Grand total' ), $total_queries_times, $total_queries_count;
```
marc-vanderwal, line 736: Again, using a single source of truth for the column widths:
```suggestion
            # Totals line
            say join " ", map { "=" x $_->{width} } @columns;
            printf "%*s %67.2f %11s\n", $max, __( 'Grand total' ), $total_queries_times, $total_queries_count;
```

marc-vanderwal (commented):
It’s fine, but some column headings in the statistics table aren’t translated. I took the liberty to suggest some little style changes as well.

lib/Zonemaster/CLI.pm
```diff
@@ -1,15 +1,14 @@
 # Brief help module to define the exception we use for early exits.
 package Zonemaster::Engine::Exception::NormalExit;
-use 5.014002;
+use v5.26;
```
matsduf, line 3: Does that mean that we increase the lowest Perl version of Zonemaster from v5.16 to v5.26?

lib/Zonemaster/CLI.pm
matsduf: The width of each the column is wide enough for the English header name, but in translation the header name might get wider. I guess the solution is to have adaptive column width for all comlumns.

Or does the second suggestion take care of that?

matsduf:
This can be merged, but three issues to be created:
* Create detailed user documentation and link to that from the script in v2025.2.
* Refactoring of the table printing code to better support translations and increase maintainability. -> https://github.com/zonemaster/zonemaster-cli/issues/439
* Refactoring of section mapping of data for both outputs. Grand total is to be included in the JSON output. -> https://github.com/zonemaster/zonemaster-cli/issues/440

matsduf (dismissed):
Conflict to be resolved.

marc-vanderwal:
> This can be merged, but three issues to be created: […]
> 
>     * Refactoring of the table printing code to better support translations and increase maintainability.

I created #439 to track that.

tgreenx:
>  [ ... ]
>     * Refactoring of section mapping of data for both outputs. Grand total is to be included in the JSON output.

I created https://github.com/zonemaster/zonemaster-cli/issues/440 for this.

matsduf (approved):

marc-vanderwal:
Successfully release tested this PR on Rocky Linux 9.

Using the command line in the test procedure, the output I obtained looks similar to the example, albeit with different numbers. The formatting of the table looks correct, though currently, none of the strings are translated.

PR #443, opened 2025-06-23
Fixes issue with locale in unit tests

### LINKED ISSUE #438: Unit tests fail on non-English system
During release testing for 2025.1, I attempted to install the latest `develop` branch of zonemaster-cli on a fresh Rocky Linux 9 VM whose default system language is French.

When running `cpanm --sudo Zonemaster-CLI-*.tar.gz`, 8 tests fail, seemingly because of the system’s locale not being English (see [build.log](https://github.com/user-attachments/files/20670469/build.log)).

When I do `LC_ALL=en_US.utf8 cpanm --sudo Zonemaster-CLI-*.tar.gz` instead, forcing the system locale to be English, only one failing test remains instead of 8 (see [build.log](https://github.com/user-attachments/files/20670508/build.log) of that attempt).

matsduf:
## Purpose

This PR fixes the issue reported in #438 and https://github.com/zonemaster/zonemaster-cli/issues/438#issuecomment-2996235684

## How to test this PR

Run the unit tests with other locale than "en" or "C" and the unit tests should pass, e.g.

* LC_ALL=da_DK.UTF-8 cpanm -v --test-only Zonemaster-CLI-*.tar.gz # Using a dist file
* LC_ALL=da_DK.UTF-8 cpanm -v --test-only Zonemaster::CLI

### DIFF
--- a/t/usage.t
+++ b/t/usage.t
@@ -19,6 +19,14 @@ use Test::Differences;
 use Zonemaster::CLI;
 use JSON::Validator;
 
+# Force locale C for these unit tests. They depend on the print outs not being
+# translated. See zonemaster/zonemaster-cli/issues/438 and
+# https://github.com/zonemaster/zonemaster-cli/issues/438#issuecomment-2996235684
+$ENV{LC_ALL} = "C.UTF-8";
+delete $ENV{LANG};
+delete $ENV{LANGUAGE};
+
+
 # CONSTANTS
 
 Readonly::Array my @SIG_NAMES => do {


marc-vanderwal (approved):
Oh, I was going to look into it but didn’t have the time to do so. Thanks!
Anyway, this looks good to me. That’s exactly the solution I was thinking of too.

tgreenx (commented):
Shouldn't we store the current values of those environment variables first and restore them at the end of the unit test? They may be in use by other programs, right?

matsduf:
> Shouldn't we store the current values of those environment variables first and restore them at the end of the unit test? They may be in use by other programs, right?

My assumption is (was) that each `t` file is run in its own process. Can a `t` file change the environment for another `t` file?

tgreenx:
> > Shouldn't we store the current values of those environment variables first and restore them at the end of the unit test? They may be in use by other programs, right?
> 
> My assumption is (was) that each `t` file is run in its own process. Can a `t` file change the environment for another `t` file?

After testing locally, it looks like it's fine yes.

tgreenx (approved):

PR #449, opened 2025-09-02
Make dist generate MO files for inclusion in the dist file.

### LINKED ISSUE #442: Translations (*.mo) files are generated inappropriately in the build process
During release testing of Zonemaster-CLI, I attempted to install the latest `develop` branch of the module using the following command line on a freshly checked-out Git repo:

```
perl Makefile.PL
make dist
cpanm --sudo -v Zonemaster-CLI-*.tar.gz
```

However, doing so causes some of the unit tests related to translations to fail. If forcing the installation, the package gets installed but the `*.mo` files are missing, meaning that none of the strings are translated into the selected locale.

The workaround is to introduce an additional step:

```
perl Makefile.PL
make
make dist
cpanm --sudo -v Zonemaster-CLI-*.tar.gz
```

This is fundamentally a problem with the build process. 

By the way, when I implement the workaround, I noticed that the resulting `Zonemaster-CLI-*.tar.gz` archive contains `.mo` files, generated from the `.po` files. It feels wrong to have generated (binary) files inside the archive; I assume that the archive should only contain source files.

mattias-p:
## Purpose

This PR updates `make dist` to generate the MO files before creating the tarball.
As a drive-by, some clean-up is also performed on `Makefile.PL`.

## Context

Fixes #442.

## Changes

* Add a dedicated make target `make -C share mo` to only generate all MO files.
* Make `make dist` depend on `make -C share mo`. Maybe there are simpler and better ways to achieve this, but the best I could come up with was:
   1. Add a new target `make mydist` that runs both `make -C share mo` and `make tardist` in that order.
   2. Override DIST_DEFAULT from `tardist` to `mydist`.
* Clean-ups in Makefile.PL:
  * Avoid reimplementing the share/Makefile mechanism
  * Use `make -C share` instead of `cd share && make`
  * Declare phony targets as such
  * Tidy Makefile.PL according to .perltidyrc

## Testing
I tested this by executing the following and verifying that all of the MO files were listed.

```sh
rm -f Zonemaster-CLI-*.tar.gz && git clean -dfx && perl Makefile.PL && make dist && clear && tar tvf Zonemaster-CLI-*.tar.gz | grep mo
```

I have not tested this on FreeBSD. My hypothesis is that the share/Makefile mechanism is able switch to gmake on FreeBSD, but I haven't actually tested it.

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -54,6 +54,11 @@ sub MY::postamble {
 
     my $docker = <<'END_DOCKER';
 
+.PHONY: mydist
+mydist:
+	$(MAKE) -C share mo
+	$(MAKE) tardist
+
 docker-build:
 	docker build --tag zonemaster/cli:local --build-arg version=$(VERSION) .
 
@@ -72,5 +77,7 @@ END_DOCKER
 
 install_script 'zonemaster-cli';
 install_share;
+
+makemaker_args( macro => { DIST_DEFAULT => 'mydist' } );
 
 WriteAll;
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -23,8 +23,8 @@ requires(
     'JSON::XS'           => 0,
     'Locale::TextDomain' => 1.23,
     'Try::Tiny'          => 0,
-    'Zonemaster::LDNS'   => 5.000000,  # v5.0.0
-    'Zonemaster::Engine' => 8.000000,  # v8.0.0
+    'Zonemaster::LDNS'   => 5.000000,    # v5.0.0
+    'Zonemaster::Engine' => 8.000000,    # v8.0.0
 );
 
 test_requires(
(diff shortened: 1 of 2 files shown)


tgreenx (approved):
LGTM. Tested and works as advertised.

@marc-vanderwal please review this solution, which should fix https://github.com/zonemaster/zonemaster-cli/issues/442
@matsduf could you test this on FreeBSD? (cf. "How to test" section)

matsduf (commented):
It works on FreeBSD, but it is not possible to run run `make distcheck` before `make all` or `make dist` has been run. I think it makes more sense to run `make all` before `make dist` to be able to run `make distcheck` before `make dist`.

Maybe `make distcheck` should be included, and `make dist` not run if distcheck fails.

matsduf (changes requested):
It is important that `make distcheck` is included in the process. It would be nice of `make dist` makes it all, but then `make distcheck` should be automatically run before the distribution file is created, and the latter should be blocked if MANIFEST is not up to date.

matsduf:
@mattias-p, what is the status of this PR?

